PR #61, opened 2015-08-31
Add button variant html files for demo.

philworthington:
Create variant html files for button demo. Update json.


Jahnp:
Approved

PR #62, opened 2015-08-31
Add/update dialog variant html files for demo.

philworthington:
Add dialog close html variant file. Update json with file order and variant files with correct class names.


Jahnp:
Approved

Issue #59, opened 2015-08-31
Small examples with screenshots


stuaxo:
All great libraries have a few small examples and screenshots visible or accessible from somewhere near the top of the README, it would be great if this had some.

andrewconnell:
@stuaxo did you check out the gulp task **watch**? It builds the whole project and creates a sample site on `http://localhost:2020` where you can see running samples.

rstormsf:
just ran `gulp watch` and I don't see any server spinning on `:2020` port :-(

rstormsf:
```
---------------------------------------------------

Fabric Message: All Fabric parts built successfully, you may now celebrate and dance!
TIP: To test changes to Fabric source, check under /samples for demo HTML files of each Component.
```
and there is no `samples` folder :-( 
also there is no log that some locahost server is running :-(

rstormsf:
I also can see there is `browserSync` dependency but it's not used anywhere. :-( 
```
gulp/fabric.js
26:var browserSync = require('browser-sync').create();
```

Jahnp:
@andrewconnell  The sample site had to be removed, at least temporarily--though there's clearly a  need for it or something similar.

In the meantime, try running `gulp`, then check under `dist/samples` to view several example pages. These include a sample form, a "video portal" mock application, an icons demo page, and individual samples for each component.

@rstormsf , oops! Good catch :) We were previously using BrowserSync for testing. We'll likely add it or a similar tool back in some way soon.

rstormsf:
check PR for the small output fix to remove dev's confusion when reading `gulp` output

stuaxo:
Ah cool - yeah I hadn't tried this - so I guess it's covered :) - was thinking of this being somewhere prominent near the beginning of the README..

ericthompson:
Agreed, @stuaxo. We'll update the README to more explicitly call this out.

andrewconnell:
This might be of interest (disclaimer, I kicked off the project, but it's open source and community driven):
http://officeuifabric.com/

More info on it:
http://www.andrewconnell.com/blog/introducing-officeuifabric-com-demos-and-reference-samples-for-office-ui-fabric

ericthompson:
Closing this due to #191 raising this to the top of the FAQ on the main README. More updates to come very soon!

PR #66, opened 2015-08-31
Add proper path for samples

rstormsf:
#59 
small doc fix in gulp logs


Jahnp:
This is awesome, thanks so much for this! One thing--and we should totally document this convention--could you actually re-open this against the develop branch? We use that one to capture pre-release changes. Thanks!

PR #69, opened 2015-08-31
Clean up component demos

philworthington:
Add variant template files and update json for components.
Label - add state variant templates.
Orgchart - component example html styles for displaying orgchart variants in demo page.
Overlay - style overlays in demo page.
Panel - add variant template files.
People picker - update persona html and classes, update variant templates.
Pivot - add variant template files, clean up pivot.less.
Toggle - update order on demo page in json.
Link - add ms-font-m class to link to specify text size.

### DIFF
--- a/src/components/Callout/Callout.less
+++ b/src/components/Callout/Callout.less
@@ -155,6 +155,10 @@
   font-size: @ms-font-size-s;
 }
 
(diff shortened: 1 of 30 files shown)


Jahnp:
Looks great! Approved

PR #70, opened 2015-09-01
Update links to CDN license for fonts and icons to a dynamic URL

Jahnp:
Update links to CDN license for fonts and icons to a dynamic URL. The actual file name of the license may change, so this helps keep the source reference resilient.


notandrew:
approved -- it works

Issue #67, opened 2015-08-31
Missing component javascripts.


dereknutile:
It seems like each of your Samples (in dist/samples/components/) are all missing their associated .js files.  For example, the ChoiceField component index.html (see here: https://github.com/OfficeDev/Office-UI-Fabric/blob/master/dist/samples/Components/ChoiceField/index.html#L15) calls a file named ChoiceField.js which doesn't exist.  Same thing for all the other components I've looked at.

Jahnp:
Great catch, thanks. I think this is actually an issue with the template used for component samples, which expects a JS file for a component even if one doesn't exist, hence the 404s. We clean that up.

dereknutile:
Awesome, go get 'em Peter.

Jahnp:
Thanks again for flagging this. I'll reactivate this issue for now until we've resolved it--we don't have an ETA yet but I want to be sure it's clear we're aware of it :)

digglek:
I am trying to make use of the dialog but the samples don't show how to open/close a dialog is there some missing js files as the sample references 

<script type="text/javascript" src="Jquery.Dialog.js"></script>
<script type="text/javascript" src="Dialog.js"></script>

However, these do not exist?

weshackett:
In the fabric site the Spinner page doesn't show the js tab alongside the demo html. It should have as it is in the markup of the page, just not showing up.

ericthompson:
@digglek - #274 should help with the JS for the Dialog. We've also included an example of a blocking dialog for extra completeness.

@weshackett - we'll get the examples up on the site asap for both the Spinner and the Dialog.

mikewheaton:
Issue #110 covers needing JS to demonstrate opening and closing a Dialog.

PR #72, opened 2015-09-01
Update index.html

iJebus:
Remove reference to NavBar.js as file does not exist.


Jahnp:
Hey @iJebus , thanks for taking the time to fix this! However, it looks like this PR is actually related to [this issue](https://github.com/OfficeDev/Office-UI-Fabric/issues/67)--long story short, it's a bug with the template we use to generate these demo files. The problem is that the HTML template expects component JS files, but not all components require JS files, hence the 404s. 

We're working on a fix in our build, and when it's ready all the demo files in `dist` will have accurate links.

PR #82, opened 2015-09-02
Simplify bower path in postinstall script.

adjohnson916:
Because npm scripts include npm bin-enabled modules in the "path" by default, we can simplify this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   },
   "private": true,
   "scripts": {
-    "postinstall": "node node_modules/bower/bin/bower install"
+    "postinstall": "bower install"
   },
   "devDependencies": {
     "bower": "^1.3.12",


Jahnp:
Ah, good catch. Though looking at it now, I realize that we likely don't need a postinstall for Bower at the moment, since we don't actually have any Bower deps; we had previously been using a few libs before trimming down for GitHub.

If you don't mind, could you go ahead and remove the scripts prop altogether? I'm happy to merge after that. Thanks!

adjohnson916:
@Jahnp Thanks for merging. I didn't get a chance to delete it. Still want that in another PR, if I have time? Or just leave it for now?

Jahnp:
@adjohnson916 If you have time and can test to make sure things are still golden after removing the line, that'd be great! I'd be happy to merge it in then. I think it might be related to (this issue)[https://github.com/OfficeDev/Office-UI-Fabric/issues/86], too, so that might be something to watch for.

Thanks again!

ericthompson:
Hey @adjohnson916 - thanks for contributing! Can you please sign the Contribution License Agreement located [here](https://cla.microsoft.com/)?

Thanks!

adjohnson916:
@ericthompson Signed CLA.

ericthompson:
Thanks, @adjohnson916!

PR #84, opened 2015-09-02
Update CONTRIBUTING with note about which branches to use

Jahnp:
Update CONTRIBUTING with note about which branches to use for pull requests.


notandrew:
looks great!

PR #85, opened 2015-09-02
fix filename case  sensitive

zhangjiayin:


Jahnp:
Great catches, thank you! I'll merge these in.

PR #87, opened 2015-09-02
Fix spacing between dropdown list items and dropdown list container

philworthington:
Move dropdown list border to pseudo element. List items now align correctly within the list container.


Jahnp:
Looks great! Approved

PR #90, opened 2015-09-03
Command Bar Fixes

battletoilet:
- Fixed search so it covers the command bar at smaller screen sizes.
- Fixed the overflow menu so it doesn't permanently hide the commands.
- Made it so the x appears all the time and not just when you type input.

### DIFF
--- a/src/components/CommandBar/CommandBar.less
+++ b/src/components/CommandBar/CommandBar.less
@@ -39,6 +39,7 @@
   padding-left: 0;
   padding-right: 15px;
   border: 0;
+  position: relative;
 
   &:focus {
       outline: none;
@@ -302,6 +303,7 @@
   display: none;
   top: 1px;
   right: 0px;
+  z-index: 10;
 }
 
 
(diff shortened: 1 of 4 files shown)


src/components/CommandBar/CommandBar.less
```diff
@@ -302,6 +303,7 @@
   display: none;
   top: 1px;
   right: 0px;
+  z-index: 10;
```
Jahnp: @ms-zIndex-front

src/components/CommandBar/CommandBar.less
```diff
@@ -334,19 +343,14 @@
...
-}
-
 
-//== Modifier: After search is active and has text, add X for clearing the box
-//
-.ms-CommandBarSearch--active.ms-CommandBarSearch--hasText {
-    & .ms-CommandBarSearch-input {
-      padding-right: @SearchBox-height;
-    }
+  & .ms-CommandBarSearch-input {
```
Jahnp: is .ms-CommandbarSearch-input supposed to be a child? & is usually used for adjoining classes/states--you might not need it here

PR #91, opened 2015-09-03
Clean up and restructure choicefield Less. Update and rename templates.

philworthington:
Rename template files and update markup for disabled choicefields.
Clean up and restructure Less.


Jahnp:
Looks really great! Approved

PR #92, opened 2015-09-04
Fixed DatePicker bug where navigating to a previous month fails

wdo3650:
Modifed the DatePicker LESS file to fix a bug where navigating using the previous button fails. Another click area was covering the previous button which caused this bug.


Jahnp:
Approved! Looks great.

PR #96, opened 2015-09-04
Fixed broken image URL

chirag64:


notandrew:
Thanks for making this change -- our team has a pending pull-request (#103) that has similar changes.  That change also includes the latest designs and updates to the presence indicator.  I'm going to close this for now.  Again, thanks for taking the time to fix this.

PR #97, opened 2015-09-04
remove protocol from CDN reference to avoid cross HTTP/HTTPS requests

andrewconnell:
While Office Add-ins will always be loaded from HTTPS, if the UI Fabric is used from a self-standing web app in HTTP, users will be error / warning when pulling across security boundaries. Remove the protocol so the browser will stick with the current page's protocol HTTP / HTTPS


Jahnp:
That's a great catch. Thanks Andrew, I'll merge this in. Sorry for the delay!

PR #98, opened 2015-09-04
Fixes #94. Fix label wrapping issue in underlined textfield.

### LINKED ISSUE #94: Broken styling for underlined & required TextField
Adding `is-required` and `ms-TextField--underlined` to the same TextField component breaks the styling

screen shot 2015-09-04 at 11 53 44

Markup for the screenshot:

```html
<div class="ms-TextField ms-TextField--underlined">
  <label class="ms-Label">Name:</label>
  <input class="ms-TextField-field" type="text">
</div>

<div class="ms-TextField is-required ms-TextField--underlined">
  <label class="ms-Label">Name:</label>
  <input class="ms-TextField-field" type="text">
</div>
```

philworthington:
Add white-space no wrap for labels in underlined variant. Clean up states in Less.

### DIFF
--- a/src/components/TextField/TextField.less
+++ b/src/components/TextField/TextField.less
@@ -9,38 +9,40 @@
 @import (reference) '../Label/Label.less';
 
 
-// Single line (default), standard (default)
 .ms-TextField {
   .ms-font-m;
   .ms-u-normalize;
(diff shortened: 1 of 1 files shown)


notandrew:
Approved - nice clean-up, too

PR #99, opened 2015-09-07
Remove Microsoft copyright from Jquery.DatePicker.js

pbjorklund:


notandrew:
The bigger issue here is that we need to separate the pickadate js from our presentational js file.  Thanks for finding this.

Jahnp:
Given that pickadate.js has been split out from Jquery.DatePicker.js as of https://github.com/OfficeDev/Office-UI-Fabric/pull/137, this change should be unnecessary now. Thanks for watching out for this, though--we appreciate the diligence!

PR #103, opened 2015-09-08
Update persona presence size. Clean up persona markup in all applicable components.

philworthington:
- Presence updated in persona component to match latest sizes in redlines. 
- Markup for persona updated - removed ms-Persona-imageCircle - unnecessary extra div between ms-Persona-imageArea and image. Less updated to account for this change for circle and square variants.
- Persona, PersonaCard, OrgChart, PeoplePicker updated with latest persona markup.


notandrew:
In just checking the samples, it looks like the descenders are getting cut off due to a small Height.

persona

philworthington:
Excellent catch. Updated and fixed.

Issue #71, opened 2015-09-01
datepicker markup too complex


kennywangjin:
to show a datepicker to a user, there are too many lines markup code.
is that possible to generate these markups via javascript, just like jquery ui datepicker do:
// input type='date'
//
//$('input[type="date"]').datepicker({});
//
this seems more convenient to use......

CuddleBunny:
+1

battletoilet:
We will definitely take a look at that. Thanks for your input.

wdo3650:
A simplified DatePicker is in the develop branch #115 .  Thanks for your input @kennywangjin.

wdo3650:
After discussing this component both internally and with the community, we believe it makes more sense to remove the jQuery-generated markup from DatePicker, since Fabric is more focused on the styling of components independent of its JS implementation. This puts the markup straight in the DOM, as it was before, allowing easier access to the elements, attributes and css without the need for Javascript to accomplish everyday tasks. 

Thank you for the feedback. It prompted a great discussion internally about about what Fabric should focus on.

PR #115, opened 2015-09-11
Simplified DatePicker Markup

wdo3650:
Simplified markup by wrapping elements inside the JQuery plugin code.


pbjorklund:
How are we supposed to set things like the input id now?

pbjorklund:
Realized now that perhaps I can get a better answer by pinging @wdo3650 directly as it's a closed issue.

wdo3650:
Thanks for the question. I assume you were you able to add id's to the input textfield before. In order to gain more information on what specifically you are looking for, can you describe what your process was and how you were using the id's?

pbjorklund:
Well I liked being able to control the name and id of the input fields in order to handle posts and using js to do things with the fields. Not sure I understand your question :)

wdo3650:
Thanks again for the bringing this up and it's a great question. PickaDate.js, which the DatePicker wraps, automatically assigns an id to the input textfield. I'm not sure how changing the id will affect the DatePicker's functionality. However, you can query the DOM to either set the id or get the id using JQuery. I would suggest getting the id as opposed to setting it in case changing the id interferes with functionality.

wdo3650:
Thanks for sharing your concerns. See issue #71 where more details about near term changes to DatePicker can be found.

PR #117, opened 2015-09-13
Removed non-existing icon "ppt-icon.png" from ListItem.less

brakmic:
I've got a WebPack error while trying to create a bundle based on fabric.components.css. The css loader from WebPack tries to load and inline images from css files but in this case it couldn't find any file named "ppt-icon.png".

### DIFF
--- a/src/components/ListItem/ListItem.less
+++ b/src/components/ListItem/ListItem.less
@@ -205,14 +205,6 @@
     color: @ms-color-neutralSecondary;
   }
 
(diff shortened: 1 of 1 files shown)


notandrew:
Actually, would you mind removing the entire class?  It was originally for demo purposes, but should have been removed.  Thanks!

brakmic:
I removed the class. 

Regards,

notandrew:
looks good.  I'll merge it shortly.
Thanks

PR #124, opened 2015-09-15
Fix: Click on navBar crashed with null reference without search box

robi26:
The NavBar plugin crashes when an item is clicked and no search box element is present.

### DIFF
--- a/src/components/NavBar/Jquery.NavBar.js
+++ b/src/components/NavBar/Jquery.NavBar.js
@@ -79,10 +79,11 @@
 
       // Hide any menus and close the search box when clicking anywhere in the document.
       $(document).on('click', 'html', function(event) {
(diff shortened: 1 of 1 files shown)


Jahnp:
Hi @robi26 , thanks so much for this and apologies for the delay in reviewing it! Sorry for the further hassle, but could you please re-open this against the `master` branch? We've [made some changes](https://github.com/OfficeDev/Office-UI-Fabric/issues/145) to our branches to simplify pull requests; `develop` has now been merged into `master`, but we need to clear out all PRs that have been opened against `develop` so we can delete it to make things even simpler. 

For the moment, I'm going to close this so we can finish #145 , but feel free to re-open it at your convenience--we'll make sure to shortlist the review at that time. Thanks again!

PR #126, opened 2015-09-15
Update download link in Getting Started

Jahnp:
Update the download link in Getting Started to explicitly point to the 1.0.0 release rather than the latest off of master. Moving forward, we'll want to make sure that this link is always updated to point to the latest release.


notandrew:
approved

Jahnp:
We'll update this link to point to the latest after we've pushed out our next release.

PR #137, opened 2015-09-16
Separated PickaDate from DatePicker

wdo3650:
This addresses #99. By removing PickaDate from the DatePicker.js source the license is retained.


notandrew:
Looks good.  I tried changing to a bunch of dates and everything looks like it's working. :)

PR #147, opened 2015-09-24
Vanilla JS Component Consistency

wdo3650:
Made the Spinner consistent with the way ProgressIndicator is being initialized in the individual template. Modified speed and size of Spinner.


wdo3650:
Reopening against Master

PR #148, opened 2015-09-24
Add presence colors to persona card

andrewconnell:
The **persona card** styles seem to only support the _available_ status, while the **persona** control supports busy, offline, dnd, away, etc.. I added these form looking at the persona control and matching the less variables. Please verify these are correct and I've added them to the right place.

Here's a codepen of these changes in action (explicit HEX colors replaced the variables):
http://codepen.io/andrewconnell/pen/NGRvxW

screen shot 2015-09-24 at 6 12 07 am

### DIFF
--- a/src/components/PersonaCard/PersonaCard.less
+++ b/src/components/PersonaCard/PersonaCard.less
@@ -272,3 +272,20 @@
     position: relative;
   }
 }
+
+// Presence colors
+.ms-PersonaCard.ms-PersonaCard--busy .ms-Persona-presence {
+  background-color: @ms-color-presence-busy;
+}
+.ms-PersonaCard.ms-PersonaCard--away .ms-Persona-presence {
+  background-color: @ms-color-presence-away;
+}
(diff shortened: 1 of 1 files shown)


philworthington:
Hey Andrew, thanks for looking into this. After reviewing the component and your codepen it looks like the way we're handling presence in the persona control can also be applied to the persona within the persona card. 

Since .ms-Persona can take a modifier to indicate presence (e.g. .ms-Persona--busy), we can actually apply that modifier directly to the .ms-Persona within the persona card as a method to indicate presence.

Check out this codepen: http://codepen.io/anon/pen/yYVYoW

Also, it looks like your persona card markup and styles are a little out of date. There were several updates made that are currently on master and should be available soon in the next release as well. 

Thanks again!

andrewconnell:
Ah… that makes sense… better to reuse the persona ones for presence. I’ll close this PR out.

The codepen’s I’m creating are for v1.0… When you guys ship a new version, I’ll tweak them for 1.1. Lots of changes since 1.0 that I haven’t incorporated (like the date picker).

PR #151, opened 2015-09-25
NavBar: Replaced overlay pseudo element with Fabric Overlay

wdo3650:
Replaced overlay pseudo element with Fabric Overlay. Fixed right alignment of menu icon.


Jahnp:
This works really great--thanks for this! Approved.

PR #153, opened 2015-09-28
Fix: Click on navBar crashed with null reference without search box

robi26:
Re-opened PR #124 because of branch reorganization.
The NavBar plugin crashes when an item is clicked and no search box element is present.


Jahnp:
This looks great! Thanks for this, @robi26 --I'll go ahead and merge this in. Cheers!

PR #154, opened 2015-09-28
V miljo/command bar updates

battletoilet:


src/components/CommandBar/CommandBar.less
```diff
@@ -53,26 +53,44 @@
...
+  @media only screen and (min-width: @ms-screen-xl-min) {
+    padding-right: 16px;
+  }
 }
 
 // Command Bar Item
 .ms-CommandBarItem {
   display: inline-block;
-  margin-right: 8px;
+  margin-right: 0;
```
philworthington: Looks like we have two margin-rights

src/components/CommandBar/CommandBar.less
```diff
@@ -128,12 +147,26 @@
   }
 }
 
+.ms-CommandBarItem.ms-CommandBarItem--hasTextOnly {
+  .ms-CommandBarItem-commandText {
+    display: inline;
+  }
+  .ms-CommandBarItem-chevronDown {
+    display: inline;
+  }
```
philworthington: Join commandText and chevronDown here

src/components/CommandBar/CommandBar.less
```diff
@@ -141,19 +174,18 @@
...
-  padding: 0 6px;
+  padding: 0 8px;
+}
 
-  @media only screen and (min-width: @ms-screen-lg-min) {
-    padding-right: 8px;
-    padding-left: 6px;
-  }
+.ms-CommandBarItem-icon {
+  font-size: 17px;
```
philworthington: @ms-font-size-l

src/components/CommandBar/CommandBar.less
```diff
@@ -128,12 +147,26 @@
...
+
 .ms-CommandBarItem-overflow {
   display: none;
 
   &.is-visible {
     display: inline-block;
   }
+
+  .ms-Icon {
+    font-size: 14px;
```
philworthington: @ms-font-size-m

src/components/CommandBar/CommandBar.less
```diff
@@ -53,26 +53,44 @@
...
 // Command Bar Item
 .ms-CommandBarItem {
   display: inline-block;
-  margin-right: 8px;
+  margin-right: 0;
   color: @ms-color-themePrimary;
   height: @CommandBarItem-height;
   outline: none;
   vertical-align: top;
+  margin-right: -4px;
```
philworthington: Looks like we have two margin-rights

philworthington:
Looking pretty good. There are a few things I noticed though:
- The transition on the searchfield looks a little slow. When reducing the browser size the 'new' button sometimes overlaps the searchfield text. Also when on mobile, the other buttons show too quickly when the searchfield is collapsing back to 50px causing some overlap.
- After entering text in the searchfield and clicking 'x', sometimes the searchfield will keep focus and stay editable and after entering text it will not go back to the 'is-active' state unless it is clicked into again. 
- When tabbing into the searchfield, the searchfield will not go to the 'is-active' state. Tabbing out will clear the entered search text.
- The ellipsis for overflow commands starts to get overlapped by the sideCommands at about 720px.

PR #160, opened 2015-09-29
Changed spinner to use linear gradient

wdo3650:
Refactored spinner to use linear gradient instead of symmetrical.


Jahnp:
Looks awesome, approved!

PR #170, opened 2015-10-02
Remove Bower dependency and postinstall script

Jahnp:
Remove Bower dependency and postinstall script from package.json. Both of these are remnants from pre-1.0 releases.


notandrew:
looks great!

Issue #118, opened 2015-09-13
when one dropdown is opened close another


gkht:
i think when you open one, then another should be closed.

Jester98x:
+1 for this.

Any update when we might see a fix?

ericthompson:
Sorry about the delay on this one - we're looking into it!

jonahkirangi:
Dropdowns now auto-close when another dropdown is opened. Resolved in pull request https://github.com/OfficeDev/Office-UI-Fabric/pull/176.

PR #176, opened 2015-10-07
Only allow one dropdown to be open at a time: Issue #118

jonahkirangi:
Resolves issue of multiple simultaneous dropdowns


notandrew:
looks good. thanks!

PR #177, opened 2015-10-07
Add truncation to PersonaCard

jonahkirangi:
Truncation added to primaryText, secondaryText, tertiaryText, and optionalText


notandrew:
looks good

PR #178, opened 2015-10-09
Fixed Datepicker sample

battletoilet:
This was really an issue with how we pulled in scripts onto the templates. The gulp task was only setup to include one script and since pickate has two it was breaking pickadate. Now the gulp task should handle an infinite number of scripts that are placed in the src/component/ folders.


gulp/fabric.js
```diff
@@ -517,7 +520,9 @@ gulp.task('component-samples-template', ['build-component-data', 'component-samp
             .on('error', onGulpError)
         .pipe(data(function () {
             var jstag = '';
-            if(typeof storedFiles[folder].script != "undefined") { jstag = storedFiles[folder].script; }
+            var files =  storedFiles[folder]["files"];
```
notandrew: there's an extra space before 'storedFiles'

gulp/fabric.js
```diff
@@ -517,7 +520,9 @@ gulp.task('component-samples-template', ['build-component-data', 'component-samp
...
+            for(var o=0; o < files.length; o++) {jstag += files[o];}
+            if(typeof storedFiles[folder][folder] != "undefined") { jstag += storedFiles[folder][folder]; }
```
notandrew: Would you mind adding some spaces to the for, if, and brackets?  Also, new lines for the jstag lines, please.  :)

notandrew:
looks good to me

wdo3650:
This looks good to me. Nice work!

PR #179, opened 2015-10-09
DatePicker Markup Revert

wdo3650:
Added DatePicker markup back into the html portion of the component.


notandrew:
looks good to me (I didn't build it, but the code looks right.)

Jahnp:
Looks great, works great! Approved.

PR #182, opened 2015-10-14
Fix for scroll on small screen devices.

ericivarsson:
ms-Dropdown-items should always have "overflow-y: scroll, not only on screen sizes above 480px.


ericthompson:
Looks good, approved!

PR #184, opened 2015-10-20
DatePicker month button fix

wdo3650:
The month button was toggling to a state that only existed on smaller screens. A media query was added to prevent this from happening on screens larger than 459px.


Jahnp:
Looks great! Tested in Chrome, Edge, and FF. Approved!

PR #186, opened 2015-10-22
suggestion to fix #185. extend the datepicker.js default values with a passed configuration 

### LINKED ISSUE #185: Configurable DatePicker
Since you're using the incredibly capable pickadate.js, why not allow configuration of that via your JQuery plugin. It's a quick fix. I'll issue a pull-request shortly and reference it here. This would make formatting, localization and customization of the Datepicker very easy.

aremes:
Makes using the datepicker much more comfortable. you can now easily set options like "firstDay", "format" etc, which pickadate.js already supports, simply by passing your own object to DatePicker().

This is just a minor patch


ericthompson:
Thanks for this @aremes! This is directly in reference to issue #185, so I'm going to go ahead and close that too.

Just to reiterate from the same issue I'm closing... the design language currently doesn't support all of the functionality in pickadate.js's robust API, so we can't take on any bugs resulting from the API's many options or the visualizations of those options.

To that end, we'll definitely look into supporting more of the functionality provided by pickadate.js (e.g. disabled days, formatting options, etc.) and we opened an issue on ourselves - #188 - to track this.

PR #194, opened 2015-11-01
Fixed a critical js bug.

vzaidman:
"ReferenceError: year is not defined at updateCustomView"


ericthompson:
Great catch. Merging!

PR #195, opened 2015-11-03
expose pickadate objects to user

vzaidman:
so the user can manipulate the input (including setting date and it's format)
also, the right version's pickadate docs should probably be linked in DatePicker's docs.

### DIFF
--- a/src/components/DatePicker/Jquery.DatePicker.js
+++ b/src/components/DatePicker/Jquery.DatePicker.js
@@ -8,7 +8,7 @@
 
   $.fn.DatePicker = function (options) {
 
-    return this.each(function () {
+    var picker = this.map(function () {
 
       /** Set up variables and run the Pickadate plugin. */
       var $datePicker = $(this);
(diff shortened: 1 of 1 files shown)


ericthompson:
Hi @vzaidman - I added a comment into the DatePicker's html file specifying the version and a link to the github documentation for that version.

wdo3650:
Thanks for the contribution @vzaidman. Being able to interface directly with the Pickadate.js API is a great idea. I noticed that the method you're using to allow users to interface with Pickadate.js eliminates JQuery methods for DatePicker e.g. chaining. This could potentially cause issues with DatePickers that are in use. There are two things I would suggest. You could create a function to return the internal Pickadate plugin. It could look something like this.

return this.each(function() {
...
    $picker = $dateField.pickadate('picker');
    $(this).data('pickadate', $picker);
...
};

Then users should be able to access Pickadate's API. For example one could gain access to the 'get' method of Pickadate to return its id:

$('.ms-DatePicker').DatePicker().data('pickadate').get("id");

Alternatively you could create public methods in DatePicker that interface with the internal Pickadate plugin.

mikewheaton:
After discussing with the team, we've decided that modifying the date picker in this way is best addressed by individual users rather than rolled into Fabric.

PR #205, opened 2015-11-11
Removed Extra ending div and added two missing divs

battletoilet:


Jahnp:
Approved, feel free to merge.

PR #206, opened 2015-11-11
- Removed slashes at the end of the a tags in command bar :(

battletoilet:


Jahnp:
Looks great, and good catch--go ahead and merge!

PR #209, opened 2015-11-15
Fixing up panel

battletoilet:
- Fixed Panel for IE 10
- Added animation classes for panel
- Added direction in Fabric animation names.
- Added open and Close classes that Javascript looks for

### DIFF
--- a/src/components/Panel/Panel.html
+++ b/src/components/Panel/Panel.html
@@ -5,7 +5,7 @@
 </button>
 
 <div class="ms-Panel">
-  <div class="ms-Overlay ms-Overlay--dark js-togglePanel ms-PanelAction-close"></div>
+  <div class="ms-Overlay ms-Overlay--dark ms-PanelAction-close"></div>
   <div class="ms-Panel-main">
         <div class="ms-Panel-commands">
             <div class="ms-CommandBar">
--- a/src/components/Panel/Panel.html
+++ b/src/components/Panel/Panel.html
@@ -1,116 +1,39 @@
 <!-- Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE in the project root for license information. -->
 
-<button class="ms-Button js-togglePanel">
+<button class="ms-Button ms-PanelAction-open">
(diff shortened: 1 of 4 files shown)


src/components/Panel/Panel.less
```diff
@@ -94,34 +135,22 @@
     border-right: 1px solid @ms-color-neutralLight;
     right: -40px;
```
philworthington: Please remove. This is moving the panel-main off the screen to the right by 40px.

src/components/Panel/Panel.html
```diff
@@ -1,116 +1,39 @@
 <!-- Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE in the project root for license information. -->
 
-<button class="ms-Button js-togglePanel">
+<button class="ms-Button ms-PanelAction-open">
```
philworthington: Maybe js-panelOpen as the classname for the js hook to make it clearer.

src/components/Panel/Panel.html
```diff
@@ -1,116 +1,39 @@
...
   <span class="ms-Button-label">Open panel</span>
 </button>
 
 <div class="ms-Panel">
-  <div class="ms-Overlay ms-Overlay--dark js-togglePanel"></div>
+  <div class="ms-Overlay ms-Overlay--dark js-togglePanel ms-PanelAction-close"></div>
```
philworthington: Looks like js-togglePanel can be removed. Classname js-panelClose.

src/components/Panel/Panel.less
```diff
@@ -5,6 +5,7 @@
 // --------------------------------------------------
 // Panel styles
 
+@ms-Panel-mdWidth: 341px;
```
philworthington: nit: should probably be 340px. Then just @ms-Panel-mdWidth + 1 down on line 138 for ms-Panel-main width if it has to be 341px.

philworthington:
Looks good and approved, just a few minor things but this definitely fixes the primary issues we were having.

PR #213, opened 2015-11-17
Changed json file to have correct component name.

wdo3650:


Jahnp:
Makes sense, good catch. Approved.

PR #225, opened 2015-12-04
Spinner redlines update with variants

wdo3650:
Updated to conform to redlines. Added large and labeled variants. Modified component example to render vanilla javascript variants.


philworthington:
Nice work, looks great! approved

Issue #203, opened 2015-11-10
SearchBox Label Incorrect Positioning


ryankmcintyre:
The label for SearchBox shows inside the search field if using Chrome, but in Edge and IE the label is shown to the right of the input field with the icon above the text label.

ericthompson:
Interesting... we're actually not getting *that exact* bug, but we're seeing that the placeholder text blocks selection of the SearchBox.

What versions of IE and Edge are you currently using?

ryankmcintyre:
Edge is 20.10240.16384.0
IE is 11.0.10240.16431 (also has "Update Versions: 11.0.22")

ericthompson:
Thanks for the extra info - we'll look into this one.

mikewheaton:
Looks like this has been fixed by @wdo3650.

PR #228, opened 2015-12-10
Fix for #203

wdo3650:
Made the whole input field clickable. Fixed bug where the close button would reappear after clicking.

### DIFF
--- a/src/components/SearchBox/Jquery.SearchBox.js
+++ b/src/components/SearchBox/Jquery.SearchBox.js
@@ -30,7 +30,7 @@
         $searchField.addClass('hovering')
       });
 
-      $searchField.on('mouseout', function(event) {
+      $searchField.on('mouseout', function() {
         $searchField.removeClass('hovering');
       });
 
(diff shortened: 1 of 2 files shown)


philworthington:
The only issue here seems to be what we were saying in scrum earlier: There's a flicker when you click the close button. Also I noticed that the event for close button is 'mousedown'. Seems like it should just be a click or mouseup. Haven't tested it but this might contribute to the flicker.

src/components/SearchBox/SearchBox.less
```diff
@@ -82,6 +89,7 @@
   display: none;
   font-size: @ms-font-size-l;
   color: @ms-color-white;
+  z-index: 2;
```
battletoilet: Use mixin @ms-zindex-front ?

src/components/SearchBox/Jquery.SearchBox.js
```diff
@@ -31,24 +40,28 @@
...
+        }, 10);
+        
         /** Only do this if no text was entered. */
         if ($(this).val().length === 0 ) {
           $(this).siblings('.ms-SearchBox-label').show();
         }
 
         // Reset cancel to false
-        cancel = false;
+        cancel = false;
```
battletoilet: Extra white space here.

battletoilet:
APPPROOVEDZZZZ

PR #231, opened 2015-12-11
Adds sass converted files to fabric

harry8525:
This adds sass converted files to allow people to use Sass with Fabric!


Jahnp:
This is unbelievably awesome to see, Harry! We'll need to make sure components are updated to SCSS before we do a full release with them, but this seriously accelerates that work. Thanks so much!

PR #234, opened 2015-12-16
Breadcrumb update

wdo3650:
Updated Breadcrumb styles and added overflow menu.

### DIFF
--- a/src/components/Breadcrumb/Breadcrumb.less
+++ b/src/components/Breadcrumb/Breadcrumb.less
@@ -14,13 +14,13 @@
       display: inline;
     }
   }
+}
 
-  .ms-Icon--chevronRight {
-    font-size: @ms-font-size-l;
-    color: @ms-color-neutralSecondary;
-    vertical-align: top;
-    margin: 10px 0;
-  }
+.ms-Breadcrumb-chevron {
+  font-size: @ms-font-size-l;
+  color: @ms-color-neutralSecondary;
+  vertical-align: top;
+  margin: 10px 0;
 }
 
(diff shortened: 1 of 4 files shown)


src/components/Breadcrumb/Breadcrumb.less
```diff
@@ -7,45 +7,165 @@
...
+}
+
+@media screen and (max-width: 480px) {
+  .ms-Breadcrumb-itemLink {
+    font-size: @ms-font-size-m;
+    max-width: 116px;
+  }
+
+  .ms-Breadcrumb {
+    .ms-Icon--chevronRight {
```
battletoilet: Maybe we should put a custom class like .ms-Breadcrumb-chevron or something just in case someone in the future wanted to use the chevron class inside the breadcrumb. Thoughts?

src/components/Breadcrumb/Breadcrumb.less
```diff
@@ -7,45 +7,165 @@
...
+  text-overflow: ellipsis;
+  overflow: hidden;
+}
+
+
+@media screen and (max-width: 640px) {
+  .ms-Breadcrumb {
+    margin: 10px 0;
+
+    .ms-Icon--chevronRight {
```
battletoilet: Same thought as comment below.

battletoilet:
Other than my comment this looks really good. I tested it locally and it works great.

battletoilet:
Actually when you tab onto the Ellipsis and hit enter, it doesn't open the contextual menu. If this is a 4+ hour task lets not worry about it but if it's simple lets try and get it in.

PR #244, opened 2015-12-31
Update content for descriptions of components from Toggle to MessageB…

ericthompson:
…anner

### DIFF
--- a/src/components/Overlay/Overlay.json
+++ b/src/components/Overlay/Overlay.json
@@ -1,6 +1,7 @@
 {
   "name": "Overlay",
   "notes": "",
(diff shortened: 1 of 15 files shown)


src/components/Overlay/Overlay.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "Overlay",
   "notes": "",
+  "description": "An element that takes over the screen to help bring focus to an element for the user. Used in conjunction with the {0} Panel {1} and {0} Dialog {1} and available in clear and dark.",
```
battletoilet: Instead of restarting the count, you have to continue the count so 
{0} Panel {1} and {0} Dialog {1} would become **{0} Panel {1} and {2} Dialog {3}**

PR #254, opened 2016-01-06
Add missing ms-Icon SASS mixin

Jahnp:
This adds a missing mixin for ms-Icon to the core SASS files that were missed in the initial conversion from LESS.


battletoilet:
Looks good, lets get it in.

PR #256, opened 2016-01-07
Updates to Segoe font face/font stack

notandrew:
* Updating font face to use local version of Segoe by default
* Adding woff2 Segoe files to the font face
* Removing Segoe WP from the font stacks since Windows Phone now uses standard Segoe
* Referencing Japanese font Yu Gothic properly now (it's Yu Gothic, not YuGothic)


mikewheaton:
The changes in this pull request look good to me. It is difficult to test localized fonts, so we may want to look into ways to do that.

PR #258, opened 2016-01-07
Add keyboard accessibility to toggle

philworthington:
- Input remains in the dom (opacity: 0)
- Tab to navigate to and focus on toggle 
- Space selects/deselects
- Add focus states


notandrew:
Tested and verified. Nice job!

PR #266, opened 2016-01-13
SearchBox: update z-index

tomaon:


notandrew:
Same as the Textfield -- can you tell us the problem you're fixing? Thanks!

wdo3650:
The issue was that the label interfered with click events on then search field.
This should be fixed in PR #294.

PR #267, opened 2016-01-13
TextField: update z-index

tomaon:


notandrew:
The change looks pretty straight-forward, but can you tell us the problem you're fixing? Thanks!

mikewheaton:
Hi @tomaon, just checking in to see if you can provide us with some details on this pull request. What problem are you solving here?

tomaon:
Same as #266, #294
Focuses label.ms-Label, not input.ms-TextField-field

wdo3650:
If it's like #266 then I'm assuming the bug here is that ms-label is preventing a click event from occurring on the input field. 

This may have been fixed in master, as I can't seem to reproduce this bug.

wdo3650:
@tomaon if I'm not mistaken this PR fixes an issue with the label preventing the caret from showing up in the input field on hover.  Nice work spotting the bug; however, the next release of Fabric, scheduled for later this week, switches over to Sass from Less. If you'd like you could file another PR with the Sass fix for this issue as Fabric will no longer be using Less.

wdo3650:
#316 has been opened to address this issue.

PR #277, opened 2016-01-16
Removes classes for icons that are no longer included with Fabric.

### LINKED ISSUE #79: Some icons are missing
Some icons are not shown in the sample page. For example ms-Icon--onedrive, ms-Icon--thumbDown, ms-Icon--thumbUp, ms-Icon--officeThreshold, ms-Icon--office365, ms-Icon--office365Preview, etc

mikewheaton:
Fixes #79

### DIFF
--- a/src/less/_Fabric.Icons.less
+++ b/src/less/_Fabric.Icons.less
@@ -40,7 +40,6 @@
 .ms-Icon--event:before { content: '\e013'; } /* '' */
 .ms-Icon--folder:before { content: '\e014'; } /* '' */
 .ms-Icon--documents:before { content: '\e015'; } /* '' */
-.ms-Icon--onedrive:before { content: '\e016'; } /* '' */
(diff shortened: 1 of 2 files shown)


notandrew:
Can you remove ms-Icon--classNotebook from the Icon sample page, too?  Thanks!

mikewheaton:
Removed. There were also "space", "thumbsUp", and "thumbsDown" icons that are no longer in Fabric. Didn't we used to generate this page with JavaScript and the JSON file generated by Fontello? That would make it a lot easier to maintain. Right now we have 339 icon classes, and only 338 icons on the sample page. It would take ages to figure out what's missing...

notandrew:
That's the plan -- we should have a workitem for that, together with auto-subsetting.

mikewheaton:
Want to make a branch for everything MDL2 related and move these classes there for now? Right now the classes are in master, but I don't think we'll want them included in the next release.

PR #279, opened 2016-01-19
Converted Components and Samples over to SASS

battletoilet:


mikewheaton:
I'm having trouble getting the build system to use Sass rather than LESS. I've run `gulp ConfigureEnvironment-setSassMode`, which looks like it should be all that's necessary to switch the build system over to Sass. I then made changes to some components' Sass files, but running `gulp` gives me the same output in fabric.components.css. What am I missing?

src/components/SearchBox/SearchBox.scss
```diff
@@ -0,0 +1,108 @@
+// Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE in the project root for license information.
```
wdo3650: Seems like this styling is from a previous version of SearchBox.

src/components/Toggle/Toggle.scss
```diff
@@ -0,0 +1,180 @@
+// Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE in the project root for license information.
```
wdo3650: Toggle behavior seems broken

mikewheaton: Check in with @wdo3650 about this, turns out he wasn't building the Sass and it works fine if you do.

mikewheaton:
I did a comparison of the dist/ output from 'master' and this branch. There are some differences (mostly in source order and things like changing `border: 0` to `border: none` but I don't see any major regressions. Should be fine to merge in once we fix the SearchBox regression @wdo3650 noticed.

PR #280, opened 2016-01-19
Grid offset classes

estruyf:
There were no offset classes that enabled you to move the elements over the grid. I have added these grid offset utility classes to the SASS file. An example of how it can be used can be tested here: http://codepen.io/anon/pen/eJeyea

### DIFF
--- a/src/sass/Fabric.Responsive.Utilities.Output.scss
+++ b/src/sass/Fabric.Responsive.Utilities.Output.scss
@@ -202,7 +202,7 @@
   left: auto;
 }
 
-// Small Offset classes 1 ... 11
+// Small Offset classes 0 ... 11
 .ms-u-smOffset11 {
   margin-left: 91.66666666666666%;
 }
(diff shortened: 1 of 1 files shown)


mikewheaton:
Thanks for this pull request, @estruyf, but we actually do already have offset classes for the grid. They are 'push' and 'pull' classes and are available for all screen sizes. For example, `ms-u-mdPush12` and `ms-u-lgPull3`. Do these meet your needs?

estruyf:
Hi @mikewheaton, you closed it a bit too fast :-). 

Offset and push/pull are not the same. Offset classes are setup with "margin-left" which will force other columns to move over the grid. It can be used to let all elements start from a specific point on the grid.

The push or pull classes are setup with "left" or "right", which does only position a specific element on which the push or pull class is defined. That also means that you could overlap some elements if you are not cautious.

I created another codepen to show you the difference: http://codepen.io/anon/pen/WrXaQm

mikewheaton:
Thanks for the new CodePen, @estruyf – that makes it much clearer how offset differs from our current push/pull. It looks like both Foundation and Bootstrap provide these offset classes (increasing the left margin) in addition to push/pull for column ordering. I can see the value here, we'll just have to discuss as a team before merging the pull request.

mikewheaton:
Looks good, thanks for contributing this!

PR #285, opened 2016-01-22
Fixes alignment of the header within a Contextual Menu when it's not …

### LINKED ISSUE #283: ContextualMenu header text out of alignment
When using list item `ms-ContextualMenu-item--header` with regular (not multi selection) menu, then header is not aligned with other menu items.  
Here the [plunk](http://plnkr.co/edit/4sTsxYedCgn5IDR5wFUU?p=preview) demonstrating the issue.

mikewheaton:
Fixes #283

Here is what a ContextualMenu now looks like with headers and not in multi-select mode:
image

### DIFF
--- a/src/components/ContextualMenu/ContextualMenu.less
+++ b/src/components/ContextualMenu/ContextualMenu.less
@@ -125,7 +125,7 @@
 //== Modifier: Multi-select menu
 //
 .ms-ContextualMenu.ms-ContextualMenu--multiselect {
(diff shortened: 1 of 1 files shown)


mikewheaton:
Also made the change in @battletoilet's 'less-sass-components' branch to be sure we don't lose this fix when that branch is merged in.

mikewheaton:
I like it!

Issue #287, opened 2016-01-23
Fix ContextualMenu animations


s-KaiNet:
ng-officeuifabric here :)  
We are implementing animation for contextual menu component. Current UI Fabric css styles make this work a little bit harder. Consider this [plunk](http://plnkr.co/edit/0mPKopikCK8RODoJ2ptH?p=preview).   
There are three menus - issue, no issue and fixed. When clicking on "issue" `Move` sub menu, fade in animation is applied correctly, but when clicking on `Delete`, during fade out there are no border (list styles are missing, etc.) on sub menu, which is not cool.   
This fixed in "no issue" menu by copying styles from UI Fabric css (consider `style.css` file and `.noissue` styles), which also is not good, cause styles may change.   

The cause of this behavior is that `.ms-ContextualMenu` (which is considered to be a "hidden" menu) and `.ms-ContextualMenu.is-open` (which is "opened" menu) have different styles. When fade out is in progress, menu has only `.ms-ContextualMenu` class styles and styles from `.ms-ContextualMenu.is-open` are missing. That's why there are no borders, etc.   
Is it better to put all required styles under  `.ms-ContextualMenu` (which is "hidden"), and "opened" menu in turns will contain only `display:block` style?  This approach demonstrated under "Fixed" menu section (file `fixed.context.menu.css`). Since `.ms-ContextualMenu` is hidden (`display:none`) it doesn't matter which styles it holds.    
From .less view it will be something like this one (`ContextualMenu.less`):   

```css
.ms-ContextualMenu {
  .ms-font-m;
  .drop-shadow;
  background-color: @ms-color-white;
  border: 1px solid @ms-color-neutralTertiaryAlt;
  list-style-type: none;
  position: absolute;
  width: 180px;
  z-index: (@ms-zIndex-ContextualMenu + @ms-zIndex-middle);
  display: none;

  // Hidden by default, then trigger this state to show it.
  &.is-open {
    display: block;
  }
}
```   
If you approve I can do a PR.

andrewconnell:
:+1:

mikewheaton:
Awesome catch, Sergei, and thanks for demonstrating the issue and solution. Please go ahead with a pull request and we'll get this fix reviewed and merged in ASAP.

PR #289, opened 2016-01-24
Fixing inconvenient styles for contextual menu.

s-KaiNet:
Solves #287

### DIFF
--- a/src/components/ContextualMenu/ContextualMenu.less
+++ b/src/components/ContextualMenu/ContextualMenu.less
@@ -8,19 +8,20 @@
 
 .ms-ContextualMenu {
   .ms-font-m;
-  .ms-u-normalize;
+  .resetMargins;
+  .resetPadding;
+  .drop-shadow;
+  background-color: @ms-color-white;
+  border: 1px solid @ms-color-neutralTertiaryAlt;
+  list-style-type: none;
+  position: absolute;
+  width: 180px;
+  z-index: (@ms-zIndex-ContextualMenu + @ms-zIndex-middle);
   display: none;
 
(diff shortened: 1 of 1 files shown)


src/components/ContextualMenu/ContextualMenu.less
```diff
@@ -8,19 +8,20 @@
 
 .ms-ContextualMenu {
   .ms-font-m;
-  .ms-u-normalize;
+  .resetMargins;
+  .resetPadding;
```
mikewheaton: Did you find another bug that required removing `ms-u-normalize` and adding `resetMargins` and `resetPadding`?

s-KaiNet: `.ms-u-normalize` contains `.resetBoxShadow` which is not desired, because menu has `.drop-shadow`. But now I see that I missed `.ms-u-borderBox` to copy from `.ms-u-normalize`. So `.ms-ContextualMenu` should contains all styles from `.ms-u-normalize` except `.resetBoxShadow`. Sorry for missing `.ms-u-borderBox`

mikewheaton: No worries, would you mind adding in `.ms-u-borderBox` back in and committing it? It makes sense to take out `.resetBoxShadow` so we should be able to merge this in once border-box is set.

s-KaiNet: Done

mikewheaton:
Looks good to me. We just need one more reviewer from our team to take a look and we can merge it in. Thank you!

Jahnp:
Looks great! Excited to merge this in.

PR #290, opened 2016-01-24
Fix the accidentally removal of selection by clicking on eclipses pivot entry

StfBauer:
This will fixe the removal of current selection when ellipses element was clicked.


mikewheaton:
Looks good. This change makes sense as the ellipses should open an overflow menu, and the active pivot would only change after a selection in that menu.

PR #293, opened 2016-01-26
add class ( is-disabled, is-required ) to ms-Toggle-description

tomaon:


mikewheaton:
We'll have to give this a quick test (particularly to make sure the toggle doesn't work when it's disabled) but this looks like a good feature to add. Thanks!

mikewheaton:
Unfortunately our switch from LESS to Sass means that we can't merge this pull request. I've created issue #315 to make sure this gets done. Thanks!

PR #294, opened 2016-01-26
Fixed #203. z-index and hover color fix

### LINKED ISSUE #203: SearchBox Label Incorrect Positioning

wdo3650:
Added correct z-index for the inputfield. In Sass used transparentize to fix hover color. Also removed shadow outline on input field focus in chrome.

### DIFF
--- a/src/components/SearchBox/SearchBox.less
+++ b/src/components/SearchBox/SearchBox.less
@@ -42,6 +42,7 @@
   position: relative;
   .ms-u-normalize;
   border: 1px solid @ms-color-themeTertiary;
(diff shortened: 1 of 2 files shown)


mikewheaton:
Looks good to me. One more reviewer and we can merge it in.

Jahnp:
Looks great--approved! Side note--it will be *so* nice when we only need to work on the SASS files :)

PR #298, opened 2016-01-28
Change NuGet package's components source folder from "src" to "dist"

Jahnp:
Previously, NuGet packages were being distributed with a `content/components` folder whose contents came from `src/components`. Instead, it should be populated from `dist/components` like the rest of the package folders.


notandrew:
Looks good to me.

mikewheaton:
Me too as well also.

PR #300, opened 2016-01-29
Typo fix in Readme.md

pra85:


mikewheaton:
Thanks for finding this! I think it's safe to merge this one in without a second reviewer.

Issue #56, opened 2015-08-31
Replace LESS with Sass


SandHilt:

donatj:
:+1:

andrewconnell:
:+1: :+1:

una:
:+1: :+1: :+1:

pascalberger:
+1

Diablo-D3:
:+1:

CuddleBunny:
I personally prefer LESS, perhaps maintaining both wouldn't be out of the question.

Jahnp:
This would be a pretty big change, but we do see the value in supporting SASS/SCSS in some way. We aren't in a position to fully replat our styles any time soon, but like @CuddleBunny mentions, I could see spinning up parallel versions at some point.

Tidbit for the sake of curiosity: we originally started with LESS because it had pretty broad adoption across product teams at Microsoft.

mrmckeb:
Agreed. Less is no longer the standard, as Sass has gotten much larger traction and a bigger following.

tidg:
👍 +1 👍

matthew-dean:
They're used by different groups for different reasons.

LESS is easier to learn as its a intentionally tight feature set to cover the vast majority of use cases, but not all.

Like CSS, LESS is declarative. SASS is imperative. So that also makes it a more familiar concept to web developers. For instance, LESS had concepts like block scoping and cascading values that SASS does not have. So it's more like CSS+ and SASS is more like PHP or Ruby with CSS-like syntax.

So LESS doesn't try to be the standard but a solution to a different set of problems as priority. We (core Less users) regularly suggest Sass when people need a more programming-language-like imperative construct. Basically, Less is the iPod and Sass is the Zune, so, yup, Sass has way more features. :-) 

Use what suits your project!

andrewconnell:
@matthew-dean by comparing SASS to Zune you have just made enemies with web developers at large! :smiley:

matthew-dean:
@andrewconnell lol a low blow, I know. But sometimes I hear "Sass is better than Less because it has more features." Which is pretty much exactly the same as saying, "Zune is better than the iPod because it has more features." Fewer features sometimes makes for faster adoption, less confusion, and easier onboarding for new devs, so that's the reason I bring up the comparison.

SandHilt:
Look this : https://css-tricks.com/sass-vs-less/

StfBauer:
[...]

PR #303, opened 2016-02-01
Remove LESS

mikewheaton:
This should get us all the way to completing #56. @battletoilet, please have a look at the build process to see if I've missed anything.


Jahnp:
Have you checked out the copy/build steps in FabricBuild.js?

mikewheaton:
Where is FabricBuild.js?

gulp/ComponentSamples.js
```diff
@@ -69,21 +69,6 @@ gulp.task('ComponentSamples-moveJS', function() {
             .pipe(gulp.dest(Config.paths.distSamples + '/Components'));
 });
 
-// Style Linting
-// ----------------------------------------------------------------------------
-
-gulp.task('ComponentSamples-styleHinting',  function() {
```
Jahnp: Interesting--is there no linting for SASS built in that we can switch to? We definitely want that included, very soon if not right now in this PR. I realize that's scope creep but it's an important aspect of development we shouldn't lose.

gulp/modules/Config.js
```diff
@@ -57,8 +54,7 @@ var Config = function() {
 		{src: this.paths.distComponents, dest: "/content/components/"},
 		{src: this.paths.distCSS, dest: "/content/css/"},
 		{src: this.paths.distJS, dest: "/content/scripts/"},
-		{src: this.paths.distLess, dest: "/content/less/"},
-        {src: this.paths.distSass, dest: "/content/sass/"}
+    {src: this.paths.distSass, dest: "/content/sass/"}
```
Jahnp: Nitpick, but mind the indentation--looks like a spaces/tabs diff issue

gulp/modules/ErrorHandling.js
```diff
@@ -186,54 +184,7 @@ var ErrorHandling = function() {
         });
     };
     /**
-     * LESSHint error handler
```
Jahnp: Same as above: even if we aren't adding SASS linting here, it's super important that we do add it shortly.

battletoilet: Adding it back right now

Jahnp:
Overall, looks really good. Biggest issue I can see is a lack of SASS linting/error reporting now that LESS is no longer part of the build at all. 

Also, this probably doesn't need to be addressed in this PR, we should consider whether we need some of the leftover config bits from supporting 2 preprocessors, e.g. running a task for setting SassMode explicitly (it's the only mode now) and the sassExtension prop in Config.js.

Jahnp:
Also, please remove component-manifest-template.less--that's no longer needed either.

PR #307, opened 2016-02-02
Remove MDL2 icons and sample

mikewheaton:
This work will be done on a separate branch of Fabric, as we won't be distributing both icon sets in a single release.


notandrew:
Looks good.

mikewheaton:
@battletoilet is taking a look now.

Jahnp:
Looks good to me!

Issue #302, opened 2016-02-01
Label consisting of multiple words wrapped on each word


waldekmastykarz:
When using a spinner with a label as showed in https://github.com/OfficeDev/Office-UI-Fabric/blob/70ad36cb77e1bbf3946cc7555381fd7e9359274c/src/components/Spinner/Spinner.Message.html in ngOfficeUIFabric (https://github.com/ngOfficeUIFabric/ng-officeuifabric/issues/151) we discovered that if the label consists of a few separate words each word is displayed on a new line rather than the label be displayed in a single line. Ideally the length of the wrapper should determine how the label wraps.

mikewheaton:
Thanks for reporting this, we'll look into a fix. For now, here is **[repro of the issue](http://codepen.io/mikewheaton/pen/XXBmZW)**.

waldekmastykarz:
:+1:

wdo3650:
This should be fixed in master.

PR #308, opened 2016-02-02
Bug #302 spinner label fix

wdo3650:
Fixed the label bug.

### DIFF
--- a/src/components/Spinner/Spinner.js
+++ b/src/components/Spinner/Spinner.js
@@ -27,6 +27,7 @@ fabric.Spinner = function(target) {
     var numCircles;
     var offsetSize;
     var fadeIncrement = 0;
+    var parentSize;
 
     /**
      * @function start - starts or restarts the animation sequence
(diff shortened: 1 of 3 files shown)


src/components/Spinner/Spinner.js
```diff
@@ -52,8 +53,18 @@ fabric.Spinner = function(target) {
...
+            spinner = document.createElement("div");
+            spinner.className = "ms-Spinner";
+            _target.appendChild(spinner);
+        } else {
+            spinner = _target;
+        }
+
         offsetSize = eightSize;
         numCircles = 8;
+        parentSize = spinner.className.indexOf("large") > -1 ? 28 : 20;
```
mikewheaton: Do we think 28px and 20px is enough of a difference? Maybe we can find examples of this in product to see what sizes are most common, it just seems a little small to me.

mikewheaton: Never mind, these are the correct sizes from the design library.

mikewheaton:
Looks good. One more reviewer and we can merge it in.

Jahnp:
Looks good!

Issue #305, opened 2016-02-01
Dropdown caret positioned wrong when dropdown has label


mikewheaton:
When dropdowns contain a label, the caret is positioned outside of the dropdown. This is caused by the caret being positioned absolutely within the `.ms-Dropdown` class, which grows when we add a label to it. Can we position it relatively to the field instead?

See **[this repro of the issue](http://codepen.io/mikewheaton/pen/OMwWgz)**.

![image](https://cloud.githubusercontent.com/assets/1382445/12735039/759e4f20-c8f9-11e5-8f94-1bce1a1cddf5.png)

Related issues that we can track here:
- Our form example does not include a dropdown.
- Our [component reference](http://dev.office.com/fabric/components/dropdown) doesn't include a label like the other form fields do.

mikewheaton:
This should be fixed once the documentation site has been updated with the latest component templates from Fabric 2.0.1. Leaving it open until we can verify the fix.

mikewheaton:
Confirmed, this positioning issue is now fixed in the latest release:

![image](https://cloud.githubusercontent.com/assets/1382445/13610926/da131684-e514-11e5-8528-fdf6f715f097.png)

PR #311, opened 2016-02-03
Fixed caret positioning for dropdown with label

estruyf:
Fixed the #305 bug with the caret that was not correctly positioned when using a dropdown with a label.

In the following Codepen you can check what had to be changed: http://codepen.io/estruyf/pen/KVxKOY
Here is a Codepen with the whole components css change (only one line): http://codepen.io/estruyf/pen/ZQMYPB


mikewheaton:
Thanks for this! Just one small request. Could you remove the first dropdown from `Dropdown.html` so that there is only the one with a label? We should always be showing these with labels.

estruyf:
Good to know, I've updated the templates.

mikewheaton:
Looks good, thanks! We need one more reviewer to take a look and then we can merge it in.

Jahnp:
Looks great to me!

PR #318, opened 2016-02-05
Jahnp/nuget updates

Jahnp:
Simplify NuGet package structure. Individual component files are removed, including their LESS/SASS, CSS, HTML, and JSON files. The "content" folder now only includes pre-compiled CSS and the core SASS files, and the sample JS under "scripts" is moved into the root of the package.


mikewheaton:
Looking good!

wdo3650:
I like what you got! Approved.

Issue #320, opened 2016-02-05
Panel header text line height too large


mikewheaton:
![image](https://cloud.githubusercontent.com/assets/1382445/12858869/7d4c93d8-cc07-11e5-90eb-f3b38b7bf370.png)

notandrew:
This issue repros with your changes.  We should probably remove the height property and add padding.

mikewheaton:
The problem with removing height is that the content area below it is absolutely positioned, so that when scrolling a panel with a lot of content the command bar and header text stay in place. So we'd have to use JS to set a top position for that content, or have the header scroll out of view.

philworthington:
The header text is within the absolutely positioned contentInner causing it to scroll out of view. This is what we've had by default and how we see panels built in other experiences. It's also safer to remove the height and add padding which I've done in my latest commit to PR #322 . Closing issue.

mikewheaton:
Fixed by PR #615. 

![image](https://cloud.githubusercontent.com/assets/1382445/15729624/ca176912-281a-11e6-980f-361e5db82fe2.png)

PR #322, opened 2016-02-05
Panel component - new variant sizes that adjust across breakpoints

philworthington:
This updates the Panel component with variant classes that adjust the panel sizes across all breakpoints. Sass file cleaned up. Variant template files cleaned up and updated.


philworthington:
Polish updates from design need to be made, hold off on review

src/components/Panel/Panel.ExtraLarge.html
```diff
@@ -1,116 +1,44 @@
...
+<button class="ms-Button ms-PanelAction-open">
   <span class="ms-Button-label">Open panel</span>
 </button>
 
-<div class="ms-Panel ms-Panel--extraLarge">
-  <div class="ms-Overlay ms-Overlay--dark js-togglePanel"></div>
+<div class="ms-Panel ms-Panel--xl">
+  <div class="ms-Overlay ms-PanelAction-close"></div>
   <div class="ms-Panel-main">
-    <div class="ms-Panel-commands">
```
battletoilet: Is there a reason for removing Command Bar?

battletoilet:
This looks pretty good, i tested it and it works great. Not sure if there is supposed to be content inside of the smallest panel. Also i left a comment about command bar but otherwise I LIKE WHAT YOU GOT.

src/components/Panel/Panel.scss
```diff
@@ -39,153 +46,103 @@ $CommandButtonBlue: #2488D8;
...
-  top: 40px; // Push it below the commands.
-  bottom: 0;
-  left: 0;
-  right: 0;
-  padding: 0 32px 20px 32px;
-  overflow-y: auto;
-}
-
-// Header text at the top of the panel.
-.ms-Panel-headerText {
```
mikewheaton: What can we do about panels with long header text? See issue #320 

If we're limiting it to one line, we should probably truncate with ellipses at the end.

mikewheaton:
`.ms-Panel-commands` and `.ms-Panel-contentInner` are set to `display: none;` in the "small, left-aligned, fixed" and "small, light dismiss, fixed" panels, so no content is visible.

notandrew:
Please see https://github.com/OfficeDev/Office-UI-Fabric/issues/320

philworthington:
@mikewheaton The small left aligned and small light dismiss panels are only to be used for Left Nav and the contextual menu on mobile respectively. The LeftNav or contextual menu would replace the contentInner and there would be no commands for this panel. This isn't apparent in the templates right now because we don't have a LeftNav component and any contextual menu updates should be a separate item.

Unfortunately we don't have a better way to view the panel in the sample page right now - using the buttons on the page to change the variants is just adding/removing classes in the default panel template, not actually using a different template and markup for each variant altogether. 

The important part was getting the variant size classes correct across breakpoints and making them available so that they could be consumed in other experiences. We should make updates to these panel templates when a leftnav and contextual menu are ready.

mikewheaton:
Sounds good, approved!

PR #324, opened 2016-02-05
Add .npmignore and update ignore in bower.json so unneeded files aren…

Jahnp:
…'t captured in distributions


mikewheaton:
Looks good. Since we have to re-publish to NPM anyway, let's include this in 2.0.1.

Issue #304, opened 2016-02-01
List and List Grid are not valid HTML 5


jbarr33:
Only **li** elements are allowed under **ul **but both List and List Grid have **div **directly under **ul**

As a workaround, tried using with bootstrap list-unstyled and surrounding the **div **with an **li **and it seems to work but the in the List Grid the short versions of the text that end with "..." don't work well since there is plenty of space but it displays just one letter of the name Ex.  "S..." when there is room for "Smith" to display.

I am trying to use fabric inside an Excel Task Pane app but I can't use it

mikewheaton:
Can you point us to where you're seeing this markup? On the [documentation page for the List](http://dev.office.com/fabric/components/list) we're not using `ul` or `li`, but just simple `div`s.

jbarr33:
On this GitHub site and if you download the source from GitHub also.
https://github.com/OfficeDev/Office-UI-Fabric/tree/master/src/components/List
List.Grid.html
List.html

On the documentation page it is 2 different ways:
-  when you click the HTML button you will not see the ul
- when you view the actual source you do see the ul  

Note that I have found using the actual source on the documentation page to be more reliable with some of the components but maybe I shouldn't do that anymore

Thanks.

mikewheaton:
Thanks for pointing this out, @jbarr33, we'll get those updated to use `div`s rather than `ul`s.

StfBauer:
@mikewheaton Honestly I would rather recommend to use proper ```UL``` / ```LI``` definition than replace the opening ```UL``` with divs. There is a reason why different HTML tags exist for different purposes. When something is named list people would expect to have a list. Otherwise HTML doesn't make sense at all W3C could have said here are your div elements and build whatever you want.
It was the same with the tables too.

On the other hand you might save a lot of additional CSS definitions by using proper element.

jbarr33:
I agree with @mikewheaton but I wasn't sure the extent to which you wanted feedback.  I think the bootstrap implementation of lists is nice and uses UL and LI.  Basically if you use UL class="list list-unstyled list-inline" in your list you can pretty much do anything to the LI that you want (including no css classes) but you also get a responsive list by default as the LI's will wrap within the container they are in.  So if you take it once step further and wrap the UL in a bootstrap grid and you had col-md-6 your list would wrap nicely within half of medium screen etc.  Bootstrap takes advantage of css inheritance a lot which results in less verbose css more times than not.

[...]

PR #331, opened 2016-02-08
Change List and ListItem markup to use <li>

mikewheaton:
See issue #304. I've updated the template for List to use `ul` and ListItem to use `li`. While a ListItem _could_ be used outside of a List as a simple `div`, for consistency I've wrapped each ListItem in a List.


Jahnp:
Fantastic--looks great! Approved.

notandrew:
I don't see any regressions.  Approved!

PR #337, opened 2016-02-09
Change opened selector for Date Picker

### LINKED ISSUE #336: Datepicker does not work with pickadate @ 3.5.6
When using latest pickadate.js (@3.5.6) there is an issue with the z-index:

[Plunkr 3.5.3 (works)](http://plnkr.co/edit/1nvSwsFmYVHj7ZYjEnBW?p=preview)
[Plunkr 3.5.6 (z-index issue)](http://plnkr.co/edit/IQGG16airKNuWHiUVoIA?p=preview)
Just click on the first datepicker.

This is caused by the --focus class not being applied, so it seems a bug in pickadate.js.
As you are including a custom pickadate.js version, I haven't raised this (yet) with pickadate.js.

mikewheaton:
The focused state has been removed in newer versions of Pickadate. This changes the CSS to target only the opened state. Fixes #336.

### DIFF
--- a/src/components/DatePicker/DatePicker.scss
+++ b/src/components/DatePicker/DatePicker.scss
@@ -56,7 +56,7 @@
 }
 
 // When a picker opens, always open it in front of other closed pickers
-.ms-DatePicker-picker--focused.ms-DatePicker-picker--opened { 
+.ms-DatePicker-picker--opened { 
   position: relative; 
   z-index: $ms-zIndex-front; 
 }


mikewheaton:
To be clear, this will not stop it from working with the version of Pickadate we currently bundle with Fabric.

Jahnp:
Makes sense. Looks good--approved.

notandrew:
Cool beans!  Approved with gusto. :fireworks:

PR #347, opened 2016-02-10
Ensures the hero button icon is centered within the circle

### LINKED ISSUE #346: Hero button icon not centered
As reported by @andrewconnell in #345, the hero button icon is not centered in the circle when using an `a` tag rather than `button`. See the screenshot below.

image

mikewheaton:
Fixes #346

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -131,6 +131,7 @@
 		font-size: $ms-font-size-s;
 		position: relative;
 		top: -8px;
+    text-align: center;
 
 		.ms-Icon {
       border-radius: 18px;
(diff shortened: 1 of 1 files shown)


wdo3650:
Looks good in all browsers except in Edge for me.

src/components/Button/Button.scss
```diff
@@ -131,6 +131,7 @@
 		font-size: $ms-font-size-s;
 		position: relative;
 		top: -8px;
+    text-align: center;
```
wdo3650: Align to the other CSS properties in this class.

mikewheaton:
The icon centers horizontally in Edge, but it sits about a pixel too low in the circle. This is more an issue with the icon font not having centered icons than a bug with this component. Even if we tweaked it for the plus icon, this component should support any icon and we can't get them all aligned.

wdo3650:
Looks good to me. Approved.

Jahnp:
Agreed--looks good.

PR #353, opened 2016-02-12
Fix for persona circle overflow bug when animating.

wdo3650:
In Chrome circle personas were briefly losing the circle effect during animations.


mikewheaton:
Looks good to me.

Jahnp:
Minor nit--we have $ms-zIndex-back for representing values of 0, but I could go either way on this one. Otherwise approved!

PR #356, opened 2016-02-13
Allow .ms-Icon to inherit line-height

### LINKED ISSUE #354: Remove line height from ms-Icon
.ms-Icon has line-height: 1. Is there a reason for that? It seems like it should inherit line-height from the parent.

mikewheaton:
Fixes #354

### DIFF
--- a/src/sass/_Fabric.Icons.scss
+++ b/src/sass/_Fabric.Icons.scss
@@ -12,7 +12,6 @@
   font-family: 'Office365Icons';
   font-style: normal;
   font-weight: normal;
-  line-height: 1;
   speak: none;
 }
 


notandrew:
Does this affect any components?

mikewheaton:
Argh, you're right, this could cause issues in the components if we're relying on line height not being inherited. I'll take off the 'review-needed' label until I've reviewed them.

I also opened issue #357 in hopes that we can automate this in the future.

mikewheaton:
As [explained in the issue](https://github.com/OfficeDev/Office-UI-Fabric/issues/354#issuecomment-184901058), we're going to wait and make this change in a future release. Closing the pull request for now.

Issue #174, opened 2015-10-06
Descenders are cut off in .ms-ListItem-tertiaryText


danyoel:
The style `.ms-ListItem-tertiaryText` cuts off descenders (the parts of a glyph below the text baseline). This reproes on Firefox 41.0.1 and IE 11.0.9600.

![descender-cutoff](https://cloud.githubusercontent.com/assets/9189744/10319099/623f2362-6c21-11e5-8c5c-6f1b0f4b5fbc.png)

Sample code:
```html
<div>
    <ul class="ms-List">
        <li class="ms-List-item">
            <div class="ms-ListItem">
                <span class="ms-ListItem-secondaryText">multimodel</span>
                <span class="ms-ListItem-tertiaryText">
                    <span>
                      This was a theme throughout the presser: 
                      New Metro funding passed last year,
                    </span>
              </span>
            </div>
        </li>
    </ul>
</div>
```

ericthompson:
Good catch - we should totally fix this. Probably because of a specifically set height. Looking into it!

PR #364, opened 2016-02-18
Updating ListItem's tertiary text so that descenders aren't obscured

notandrew:
Removing line-height and adding negative margins to not obscure the descenders.

Fixes https://github.com/OfficeDev/Office-UI-Fabric/issues/174


mikewheaton:
I saw you test this with PerfectPixel, and it looks pixel perfect.

Jahnp:
Yep! Just tested myself--looks great.

Issue #192, opened 2015-10-30
Update table markup


cbarnes01:
How do you accomplish column spans with the table component?

jigargandhi:
+1

StfBauer:
The table implementation is in general not recommended. You can use the same css classes and assign it to the table related HTML Elements.
Column span can only be applied directly on td and th elements but there is no equivalent to columns span available in CSS.
This will never work with the current implementation.

mikewheaton:
Our [table component](http://dev.office.com/fabric/components/table) was originally built for use in a scenario where the layout had to change completely on smaller screen sizes. This resulted in the use of `div`s rather than a standard `table` element.

We should revisit this decision, as using `table` would simplify the HTML and allow it to behave as you'd expect -- such as by respecting the `colspan` and `rowspan` attributes.

StfBauer:
The same effect could be achieved with tables too because they are only effected by the display properties:
table ..... display: table
tr ........... display: table-row
td .......... display: table-cell

Those properties can always be changed to other display behaviors like block, inline-block, inline and so on. While there was a lack on browser support on older browser using div based table looked like to be good an alternative to tables.
Alternatively a table can always be embedded in a div element that allows scrolling through overflow.
The benefit is that the data kept together but doesn't break the mobile design.

mikewheaton:
Over in #276, @StfBauer demonstrates that our current classes can be used with a standard `table` structure.

@cbarnes01: You should be able to replace the `div`s in your current implementation with `table`, `td`, `tr`, etc. and then use the [colspan attribute](http://www.w3schools.com/tags/att_td_colspan.asp) to get the result you're looking for.

StfBauer:
@mikewheaton From my side a blog post concerning this issue is coming up tomorrow. How responsive pattern can be applied to the table or at least in general how you can deal with table it in responsive scenarios.

StfBauer:
@mikewheaton Spotted and documented some issues and maybe improvements.

mikewheaton:
Excellent blog post, @StfBauer! We'll be using it as a reference as we update the reference HTML and consider responsive behaviors.

mikewheaton:
Okay, we're going to go ahead and change the markup from `div`s to a `table` element. I've opened a [new issue](https://github.com/OfficeDev/Office-UI-Fabric/issues/292) to get design input on making tables responsive.

PR #367, opened 2016-02-19
Allow table markup to be used to render Fabric table

notandrew:
Looks like the biggest difference between the table structure and the current implementation is the border-collapse property needs to be set for tables.
Converting sample markup from divs to tables.

Note: leaving 'display: table' and 'display: table-cell' properties so that users can choose tables or divs.  

Fixes: https://github.com/OfficeDev/Office-UI-Fabric/issues/192

### DIFF
--- a/src/components/Table/Table.scss
+++ b/src/components/Table/Table.scss
@@ -5,14 +5,13 @@
 // --------------------------------------------------
 // Data table styles
 
-table,
 .ms-Table {
   display: table;
   width: 100%;
   border-collapse: collapse;
 }
 
-tr,
+.ms-Table tr,
 .ms-Table-row {
   display: table-row;
   height: 30px;
@@ -54,23 +53,23 @@ tr,
   }
 }
 
-th,
-td,
+.ms-Table th,
+.ms-Table td,
 .ms-Table-cell {
   display: table-cell;
   padding: 0 10px;
 }
 
 // Style the first row as a header.
-thead td,
-thead th,
+.ms-Table thead td,
+.ms-Table thead th,
 .ms-Table-head {
   font-family: $ms-font-family-semilight;
   font-size: $ms-font-size-xs;
   color: $ms-color-neutralSecondary;
 }
 
-thead,
+.ms-Table thead,
 .ms-Table-head {
   
   td,
@@ -94,7 +93,7 @@ thead,
 }
 
 // On selectable tables, each row has a checkbox.
-.ms-Table-rowCheck {
+.ms-Table .ms-Table-rowCheck {
   display: table-cell;
   width: 20px;
   position: relative;
--- a/src/components/Table/Table.scss
+++ b/src/components/Table/Table.scss
@@ -61,7 +61,6 @@
 }
 
 // Style the first row as a header.
-.ms-Table thead td,
 .ms-Table thead th,
 .ms-Table-head {
   font-family: $ms-font-family-semilight;
@@ -93,7 +92,7 @@
 }
 
 // On selectable tables, each row has a checkbox.
-.ms-Table .ms-Table-rowCheck {
+.ms-Table-rowCheck {
   display: table-cell;
   width: 20px;
   position: relative;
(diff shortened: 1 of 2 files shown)


mikewheaton:
Nice!

What do you think about making the classes `.ms-Table-head`, `.ms-Table-row`, and .`.ms-Table-cell` optional when used inside a `<table class="ms-Table">`? We don't usually want to tie our styles to any one HTML structure, but in this case those are the only elements you'd find in a `<table>` anyway. Foundation and Bootstrap do something similar and it cuts out a lot of redundant markup.

notandrew:
That's a good point -- I'll add table, tr, and td properties too.

src/components/Table/Table.scss
```diff
@@ -5,12 +5,14 @@
...
 
-
+table,
 .ms-Table {
   display: table;
   width: 100%;
+  border-collapse: collapse;
 }
 
+tr,
```
Jahnp: Styling all of these elements makes sense, but we should maintain the opt-in spirit of Fabric by prefixing the selectors with .ms-Table, e.g. `ms-Table tr`.

src/components/Table/Table.scss
```diff
@@ -52,18 +53,31 @@
...
+.ms-Table th,
+.ms-Table td,
 .ms-Table-cell {
   display: table-cell;
   padding: 0 10px;
 }
 
 // Style the first row as a header.
-.ms-Table-row:first-child {
+.ms-Table thead td,
```
mikewheaton: Would there be a `td` inside the `thead`?

src/components/Table/Table.scss
```diff
@@ -79,11 +93,12 @@
 }
 
 // On selectable tables, each row has a checkbox.
-.ms-Table-rowCheck {
+.ms-Table .ms-Table-rowCheck {
```
mikewheaton: Don't need the `.ms-Table` selector here, just `.ms-Table-rowCheck`.

src/components/Table/Table.scss
notandrew: There can be, but it's best to use th in thead.  I'd be okay removing this line.

src/components/Table/Table.scss
```diff
@@ -52,18 +53,30 @@
...
+.ms-Table-head {
   font-family: $ms-font-family-semilight;
   font-size: $ms-font-size-xs;
   color: $ms-color-neutralSecondary;
+}
 
-  .ms-Table-cell, .ms-Table-rowCheck {
+.ms-Table thead,
+.ms-Table-head {
+
```
philworthington: nit: remove empty space

src/components/Table/Table.scss
```diff
@@ -5,12 +5,13 @@
...
 .ms-Table {
   display: table;
   width: 100%;
+  border-collapse: collapse;
 }
 
+.ms-Table tr,
 .ms-Table-row {
   display: table-row;
   height: 30px;
```
philworthington: Looks like line-height being set to 30px is making the computed row height 31px. I'm not sure if we need the line-height in there but if not, we should remove it.

src/components/Table/Table.scss
```diff
@@ -5,12 +5,13 @@
...
 // Data table styles
 
-
 .ms-Table {
   display: table;
   width: 100%;
+  border-collapse: collapse;
 }
 
+.ms-Table tr,
```
philworthington: Looks like line-height being set to 30px is making the computed row height 31px. I'm not sure if we need the line-height in there but if not, we should remove it.

philworthington:
Almost too good. Only a couple nits - one of them is noted above and another is on line 47 and 87 - we could be using $ms-font-s. I'm unable to check or select an item in the table on the demo page as well. It also looks like line-height is causing the row to be 31px instead of 30px. Other than that, this looks good.

src/components/Table/Table.scss
notandrew: done

src/components/Table/Table.scss
notandrew: I'm removing height but keeping line-height...

PR #368, opened 2016-02-19
Updated out of date dependencies, fix List style type

battletoilet:
Updated them, yay! I also fixed a strange issue with List where it was showing the list-style-stype


mikewheaton:
Looks good to me.

Jahnp:
Looks good to me as well! Approved :)

It is worth noting that csssnano uses aggressive optimizations by default (see the readme on the [original module's repo](https://github.com/ben-eb/cssnano)), which can actually merge common selectors and rebase z-indices. As a result, .min.css files can have unexpectedly different output from the sources. I think this is ok, since it actually saves an appreciable amount of space and seems to be the recommended path for minification, but worth noting in case any discrepancies related to these changes come up.

battletoilet:
@Jahnp Good point, i'm actually going to spend some more time testing before i pull this in.

battletoilet:
Alright i tested out everything i could, styles, colors, type, components using the minified version here but haven't found any issues so i am going pull this in.

PR #387, opened 2016-02-24
Fix bug with DatePicker where selecting a month would lock up.

dammittom:


mikewheaton:
Looks good to me.

Jahnp:
Looks great! Approved.

PR #394, opened 2016-02-29
Removes 'href' attribute from sample

### LINKED ISSUE #379: Remove href from Breadcrumb
In HTML5 the `href` [attribute is optional](https://github.com/OfficeDev/Office-UI-Fabric/issues/378). These should be removed from Breadcrumb, so that clicking does not cause the page to scroll. This is the same change we made to CommandBar.

mikewheaton:
Fixes #379.

Removes the 'href' attribute from samples to prevent unwanted scrolling when clicking a breadcrumb item. Updates the JS to only add the 'href' attribute to generated links if a value was provided. Ensures that the cursor is a pointer on hover.

### DIFF
--- a/src/components/Breadcrumb/Breadcrumb.html
+++ b/src/components/Breadcrumb/Breadcrumb.html
@@ -10,27 +10,27 @@
   </div>
   <ul class="ms-Breadcrumb-list">
     <li class="ms-Breadcrumb-listItem">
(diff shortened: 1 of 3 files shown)


notandrew:
approved :cake:

philworthington:
Nice, approved

PR #409, opened 2016-03-04
Build update for Markdownfiles

battletoilet:
- Changed the build to work with the new Markdown structure
- Replaced placeholder tokens for fileinclude components
- Added missing Markdown files

We still need to style up the new markdown pages but if you look at the demo, you can now see the markdown file converted to HTML and in the dist directory.


mikewheaton:
Beautiful, thanks John!

PR #427, opened 2016-03-10
Sets a max-height on the dropdown list items

### LINKED ISSUE #329: Dropdown should be scrollable
The dropdown currently has no max-height or scroll behavior. If the list of options is long, it makes the document longer and requires scrolling out of context.

image

mikewheaton:
This causes it to scroll if the list has more than 5 items.

Fixes #329.


notandrew:
Verified it works as expected - nice.

wdo3650:
Looks good to me. Nice work!

PR #430, opened 2016-03-12
Thocon/typescript components

dammittom:
# TypeScript conversion

This pull request adds TypeScript support to our build tools, and converts the component source code to TypeScript. This is setting the framework for general component code cleanup in the future, with the primary objective to remove jQuery and more clearly separate operational and presentational JavaScript.

# Build Notes

This moves all of the JavaScript related gulp tasks into the ComponentJS.js file. Also, the samples html files for each component now includes the concatenated fabric .js file, instead of a standalone .js file with just the single component.

# Usage Notes

All of the code for the Fabric components are now under the Fabric namespace. The ones that are using jQuery are no longer implemented as jQuery plugins. The next step on our roadmap is to remove jQuery completely, which may be done in increments, so it was our goal to make this breaking change for all components all at once.

For example, any Fabric Dialog components in the DOM would be initialized this way now:
```js
var elements = document.querySelectorAll('.ms-Dialog');
var i = elements.length;
var component;
while(i--) {
  component = new fabric['Dialog'](elements[i]);
}
```


Jahnp:
This is awesome! Great work--I'm super excited to dig in here.

src/components/Breadcrumb/Breadcrumb.ts
```diff
@@ -0,0 +1,235 @@
...
+
+    /**
+     *
+     * @param {HTMLElement} container - the target container for an instance of Breadcrumb
+     * @constructor
+     *
+     * If dynamically populating a list run the constructor after the list has been populated
+     * in the DOM.
+     */
+    constructor(container: any) {
```
wdo3650: Move the constructor to the top, just after the declaration of the class variables.

wdo3650:
Thoughts on providing return types for functions and typing arguments as well as function scoped variables? The advantage is it offers errors in IDE's and build tools in the event a wrong type is used. Also code can be more easily understood and debugged.

src/components/Breadcrumb/Breadcrumb.ts
```diff
@@ -0,0 +1,234 @@
...
+
+  /**
+   * Breadcrumb component
+   *
+   * Shows the user"s current location in a hierarchy and provides a means of navigating upward.
+   *
+   */
+  export class Breadcrumb {
+    // medium breakpoint
+    private static MEDIUM: number = 639;
```
wdo3650: const could be better than private static.

src/components/CommandBar/Jquery.CommandBar.ts
```diff
@@ -0,0 +1,178 @@
...
+
+      // Go through process and save commands
+      allCommands = this._saveCommands($CommandBarItems, $OverflowCommandWidth, $CommandMainArea);
+
+      // Initiate process commands and add commands to overflow on load
+      overFlowCommands = this._processCommands(allCommands, $CommandMainArea.innerWidth(), $OverflowCommandWidth);
+      this._processOverflow(overFlowCommands, $OverflowCommand, $OverflowMenu);
+
+      // Set Search Behavior
+      if ($(window).width() < 640) {
```
wdo3650: Could this be a constant?

src/components/Breadcrumb/Breadcrumb.ts
```diff
@@ -0,0 +1,234 @@
...
+/**
+ * @namespace fabric
+ */
+namespace fabric {
+  "use strict";
+
+  /**
+   * Breadcrumb component
+   *
+   * Shows the user"s current location in a hierarchy and provides a means of navigating upward.
```
notandrew: sp: user"s

src/components/SearchBox/Jquery.SearchBox.ts
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE in the project root for license information.
+// @TODO - we can add this once jquery is removed
```
notandrew: Can you file a bug with this comment (it can be generic -- enable 'use strict' and remove TODO comments)

src/components/Breadcrumb/Breadcrumb.ts
wdo3650: Actually this would need to be outside of the class implementation to work as a constant which creates other issues. So we can leave this as is.

src/components/CommandBar/Jquery.CommandBar.ts
wdo3650: Nope. See above.

dammittom:
Canceling this PR, as this will be merged over in the 3.0 branch instead.

Issue #391, opened 2016-02-27
Animate the Toggle


mikewheaton:
Right now the Toggle jumps instantly from one side to the other. This should be animated so that it's a less jarring transition.

mikewheaton:
We're going to use our existing animation classes to animate the change in position.

mikewheaton:
The toggle is now animated.

PR #431, opened 2016-03-14
Issue #391, added toggle animation

chrisahhh:
Added toggle animation using transform + transition
Breaks backwards compatibility, the toggle handle is now :after rather :before


src/components/Toggle/Toggle.scss
```diff
@@ -76,36 +79,59 @@
...
   &:before {
+    content: '';
+    width: 0;
+    position: absolute;
+    height: 16px;
+    top: 0;
+    left: 0;
+    background-color: $ms-color-themeDark;
+    outline: 2px solid transparent;
+    transition: width 0.3s;
```
mikewheaton: Could we increase the speed to 0.167s and use one of the easing curves from our [current animation classes](http://dev.office.com/fabric/styles#animations)?

mikewheaton:
@philworthington Could we use the Toggle animation from the MDL2 branch on our current Toggle control?

philworthington:
Yeah we most likely need guidance around things like animation for the toggle. The mdl2 branch includes an animation inspired by the one used for checkboxes also. There's been no input from design as to what these animations should actually look like however.

mikewheaton:
@chrisahhh Could you merge the latest master into your branch to resolve the merge conflicts? If we can test it and the animation works I think we'll go ahead and merge, although we will likely need to tweak things later when we have more design input here.

mikewheaton:
Unfortunately we can't merge this PR. Given that the Toggle animation has already been done in the updated design language (to be released in Fabric 4.0) I'm going to close this. Thank you for the contribution – if you'd like to open a new PR based on the current master we can most likely merge it.

PR #432, opened 2016-03-14
Fix #424, spinner no longer speeds up when restarted

### LINKED ISSUE #424: Spinner start() should have no effect if already spinning
If you go to this [Plunker of the Spinner](http://plnkr.co/edit/j5M6EseO6DNmEbwSIdK3?p=preview) and click on "Start spinner" multiple times, the spinner goes faster and faster and the "Stop spinner' button no longer works. Have I done something incorrect in the setup, or is this a bug?

chrisahhh:
If `setInterval` is called multiple times then there will be multiple intervals animating the spinner, which makes it look faster

Added clearInterval, to clear the previous interval

### DIFF
--- a/src/components/Spinner/Spinner.js
+++ b/src/components/Spinner/Spinner.js
@@ -34,6 +34,7 @@ fabric.Spinner = function(target) {
      * @memberOf fabric.Spinner
      */
     function start() {
+        clearInterval(interval);
         interval = setInterval(function() {
             var i = circleObjects.length;
             while(i--) {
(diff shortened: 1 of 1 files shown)


src/components/Spinner/Spinner.js
```diff
@@ -34,6 +34,7 @@ fabric.Spinner = function(target) {
      * @memberOf fabric.Spinner
      */
     function start() {
+        clearInterval(interval);
```
wdo3650: There's a stop() method in the code you can call that clears this interval. That would be great if you could swap this clearInterval out and call stop() in its place.

wdo3650:
Thanks @chrisahh, this looks great! Approved.

Jahnp:
Looks awesome! Thanks for fixing this, @chrisahhh :) We'll go ahead and merge this in.

PR #433, opened 2016-03-14
Fix #366, command bar link no longer underlined

### LINKED ISSUE #366: CommandBarItem underlines links
As pointed out by Paul Schaeflein in [Office UI Fabric - first impressions](http://www.schaeflein.net/office-ui-fabric-first-impressions/), using an `a` element in a CommandBarItem results in the link being underlined:

image

We need `text-decoration: none;` here.

chrisahhh:
Icon color fixed
Link no longer underlined

Fixes #366


mikewheaton:
Thanks for the PR! Looks good to me, we just need one more review to merge it in.

notandrew:
Yep -- looks good to me too.  Thanks

PR #434, opened 2016-03-14
Fix #345, hero button react whitespacing fix

### LINKED ISSUE #345: Hero button relies on space in HTML
Here is what the primary button looks like when rendered with React:
image

The lack of space between the plus icon and text is caused by React stripping out spaces between HTML elements. We rely on that single space to push these apart, as both the icon and text are `display: inline-block`. Need to position these in a way that will work with or without a space in the HTML.

chrisahhh:
Change button font size to 0
Add margin-left to neighbouring children

Fixes #345

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -124,6 +124,11 @@
   border: none;
   vertical-align: top;
   line-height: normal;
+  font-size: 0;
+
+  > * + * {
+    margin-left: 4px;
+  }
 
   .ms-Button-icon {
     color: $ms-color-themePrimary;


src/components/Button/Button.scss
```diff
@@ -124,6 +124,11 @@
   border: none;
   vertical-align: top;
   line-height: normal;
+  font-size: 0;
+
+  > * + * {
```
mikewheaton: I'm sure this works, but I'd prefer we use class selectors to keep the code readable.

src/components/Button/Button.scss
```diff
@@ -124,6 +124,11 @@
   border: none;
   vertical-align: top;
   line-height: normal;
+  font-size: 0;
```
mikewheaton: There seem to be some [potential issues with this method](https://css-tricks.com/fighting-the-space-between-inline-block-elements/#article-header-id-3).

src/components/Button/Button.scss
chrisahhh: I think the options are either `* + *` which covers every scenario including elements non-standard elements that someone might want to add or a big selector such as `span + span, .ms-Button-icon + .ms-Button-description, .ms-Button-description + .ms-Button-icon, .ms-Button-icon + .ms-Button-text, .ms-Button-text + .ms-Button-icon...` 

I agree `* + *` is a horrible selector

src/components/Button/Button.scss
chrisahhh: I don't know of any better solution, from the looks of it on android it fails gracefully and shows the white space

mikewheaton:
I've opened a new pull request, #652, which uses flexbox to solve this problem. It seems to be the most robust and simple solution now that we no longer support IE9. Thank for your contribution and discussion, @chrisahhh

PR #435, opened 2016-03-14
Remove underline from anchor button

chrisahhh:
Set text decoration none for a.ms-Button


src/components/Button/Button.scss
```diff
@@ -63,6 +63,10 @@
 	}
 }
 
+a.ms-Button {
```
mikewheaton: It should also remove it if we add `text-decoration: none;` directly to `.ms-Button`, right? We try to avoid targeting specific elements wherever possible.

Jahnp:
Hey @chrisahhh, thanks so much for your contribution! Just wanted to follow up on this PR, because we'd like to include the fix in an upcoming release of Fabric. Do you think you could address Mike's feedback when convenient? We're happy to merge this in after that adjustment has been made. Thanks!

mikewheaton:
This issue was fixed as part of our work on Fabric 3.0, which just merged into master.

image

Issue #446, opened 2016-03-19
Pivot: Missing functionality from component


av-virlan:
Currently pivots are able to switch styles between the pivot-links but do not hide/show any content. The way I see it Pivots should act like Bootstrap tabs and allow to to link them to some elements on the page that should be visible/hidden depending on what pivot you select.

I have an implementation in place - see [commit](https://github.com/av-virlan/Office-UI-Fabric/commit/cf2f38d6cbe4972834861af7f1db6da5582829c2) - and will prepare a pull-request if you think this is something that should be included.

mikewheaton:
The Pivot component now includes content areas that are shown/hidden based on the selected item:

![image](https://cloud.githubusercontent.com/assets/1382445/15797928/606d9bbc-29dc-11e6-8b80-b61533a5047d.png)

PR #447, opened 2016-03-19
Added pivot functionality similar to Bootstrap tabs

av-virlan:
#446 
Implemented support for pivots to be attached to a content element. Currently pivots are able to switch styles between the pivot-links but do not hide/show any content. 

**Commit details**
1. I've updated the pivot JS code to support switching between show/hide for attached pivot-content.
2. I've defined the classes for `ms-Pivot-content`
3. I've updated the pivot samples to include the pivot contents

**Implementation details**
1. Used generic selectors for the data attribute which says what the pivot-link points to so that people aren't forced to use IDs or classes for their content elements
2. Introduced the `ms-Pivot-content` CSS class only as a wrapper on top of the CSS display property. Similar to the Panel css

### DIFF
--- a/src/components/Pivot/Pivot.Large.html
+++ b/src/components/Pivot/Pivot.Large.html
@@ -1,10 +1,37 @@
 <!-- Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE in the project root for license information. -->
 
 <ul class="ms-Pivot ms-Pivot--large">
-  <li class="ms-Pivot-link is-selected">My files</li>
-  <li class="ms-Pivot-link">Recent</li>
-  <li class="ms-Pivot-link">Shared with me</li>
-  <li class="ms-Pivot-link ms-Pivot-link--overflow">
-    <i class="ms-Pivot-ellipsis ms-Icon ms-Icon--ellipsis"></i>
-  </li>
+    <li class="ms-Pivot-link is-selected" data-pivotContent="[data-demo='large']>.files">My files</li>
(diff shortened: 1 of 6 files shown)


mikewheaton:
Thanks for this contribution, @av-virlan! As you've seen, the current JS for the Pivot is simply there to demonstrate how the state classes should be toggled and isn't of much use in an actual app. We'll review this pull request and get back to you if any changes are necessary.

mikewheaton:
Looks good to me.

HeToC:
So whats the point of creating pivot tab headers without implementing tab functionality ?

Do you erally think we're interested in "how state classes should be toggled" instead of creating apps using your frabework and contributing to it ?

mikewheaton:
@HeToC: That's what this pull request adds – functional pivots. The original JS was to demonstrate the classes that are available in Fabric and not intended for production use. We're very interested in taking whatever contributions you'd like to offer. :smile:

HeToC:
@mikewheaton Actually i'd like to make office ui fabric react compatible without jquery reference, should i create new repository or create pull requests here ?
ps im new to opensource so sry for silly qs ;)

mikewheaton:
@HeToC: This core Fabric project is focused on HTML and CSS. Any implementation of these components in another framework should be done in a separate repo. You can look at [ngOfficeUIFabric](https://github.com/ngOfficeUIFabric/ng-officeuifabric) to see an implementation in Angular and [react-fabric](https://github.com/kmees/react-fabric) to see one in React.

PR #448, opened 2016-03-21
Add base i18n (internationalization) mixins

Jahnp:
# Internationalization mixins
This pull request add base internationalization (aka i18n) mixins for use in components, samples, and general use. They allow directional CSS properties--such as `left`, `right,` `border-left`, `border-right`, etc.--to appropriately change their values based on the document's `dir` attribute. The values are automatically generated in the output of SCSS files, rather than requiring a separate build step & output file. In other words, they can be used to generate LTR and RTL styles within a single CSS file.

## Usage 

To use these mixins, simply `@include` the mixin for the property you wish to use, and pass in the value you wish to have flipped as an argument. For example, generating left-to-right and right-to-left `text-align` on a `.TextField-field` element might look like this:

```scss
@import '_Fabric.Mixins.RTL.scss`;

.TextField-field {
  @include text-align(right)
}
```
This will generate the following CSS:

```css
/* Left-to-right styles. */
html[dir='ltr'] .TextField-field {
  text-align: right;
}

/* Right-to-left styles. */
html[dir='rtl'] .TextField-field {
  text-align: left;
}
```

Notice that styles for both left-to-right and right-to-left documents are generated.

## Notes

This pull request does not change any of the current Fabric components or styles to use these mixins. That is, `fabric.rtl.css` is still the primary way to achieve RTL styles in your add-in or web app using Fabric for now. That work will likely happen in a future release, but does not have a release or release date planned quite yet.


src/sass/_Fabric.i18n.scss
```diff
@@ -0,0 +1,315 @@
...
+
+//
+// Office UI Fabric
+// --------------------------------------------------
+// Internationalization mixins
+
+
+// Base left-to-right mixin definition.
+@mixin LTR {
+  html[dir='ltr'] & {
```
mikewheaton: Why not display left-to-right by default, and then override with right-to-left styles? Requiring a `dir` attribute for the styles to be applied is a big change.

src/sass/_Fabric.i18n.scss
```diff
@@ -0,0 +1,315 @@
...
+    @content;
+  }
+}
+
+// Base right-to-left mixin definition.
+@mixin RTL {
+  html[dir='rtl'] & {
```
mikewheaton: The `dir` attribute can be applied to any element, not just `html`.

src/sass/Fabric.scss
```diff
@@ -9,6 +9,7 @@
 @import './Fabric.Utilities.Output';
 @import './Fabric.ZIndex.Variables';
 @import './Fabric.Mixins';
+@import './Fabric.i18n';
```
mikewheaton: We have other files with internationalization/localization mixins. Maybe we should namespace these (e.g. `Fabric.Mixins.Localization.RTL.scss`) or place them in subfolders.

src/sass/_Fabric.i18n.scss
Jahnp: Good point. I've removed the `html` requirement.

src/sass/_Fabric.i18n.scss
Jahnp: The overrides can become a little tricky in some specific inheritance scenarios, but I agree. Following up with some eng partners who helped make this in the first place.

src/sass/Fabric.scss
Jahnp: I like the idea of namespacing these, and changing the name to `RTL`. i18n, while fun to say, is not terribly intuitive. I think longer-term we'll want to move partials into a `partials` subfolder, but for now I'll rename it to `_Fabric.Mixins.RTL.scss`.

mikewheaton:
Looks good! It's a bit unusual to check in mixins that we're not using, but we will be soon and I know some projects are eagerly waiting for these. Fabric 2.2.0 will be published today (it's been in draft for over a week) so I'll squeeze it in.

PR #449, opened 2016-03-21
Add check for selecting option from the dropdown

MIchaelMainer:
Line [133](https://github.com/OfficeDev/Office-UI-Fabric/compare/master...MIchaelMainer:master#diff-65dddecff5527031b915a8511524f674L133) -- added !$(this).hasClass('is-disabled') to check that the currently selected option is not disabled. Current behavior is the disabled option raises a change event. Updated behavior is to check that the option is disabled and not raise a change event on a disabled option.

Trimmed whitespace


mikewheaton:
Good idea, we shouldn't fire an event if the option is disabled. Looks good to me. One more approver and we can merge it in.

Issue #462, opened 2016-03-26
Spinner: Circles not always round


waldekmastykarz:
If you add:

```html
<h1 class="ms-font-xxl">
```

just before the spinner, eg.

```html
<div id="componentWrapper">
  <div class="sample-wrapper">
    <h1 class="ms-font-xxl">Hello world</h1>
    <div class="ms-Spinner"></div>
  </div>
</div>
```

circles inside the spinner won't be round anymore:
![screen shot 2016-03-26 at 22 23 25](https://cloud.githubusercontent.com/assets/11164679/14062505/861aadf8-f3a1-11e5-84bb-786e4992328a.png)

If you either use a `div` instead of `h1` or use `h1` without the `ms-font-xxl` class the spinner is rendered correctly:
![screen shot 2016-03-26 at 22 23 49](https://cloud.githubusercontent.com/assets/11164679/14062506/861af8ee-f3a1-11e5-9910-ce525f7f0341.png)
![screen shot 2016-03-26 at 22 24 03](https://cloud.githubusercontent.com/assets/11164679/14062504/85f6765e-f3a1-11e5-92a4-c8a1f0d3e26d.png)

I've noticed the same behavior in the latest versions of Chrome and Safari on OS X.

mikewheaton:
Well this is bizarre. Here is a [repro of the issue](http://plnkr.co/edit/qqfsll?p=preview):

![image](https://cloud.githubusercontent.com/assets/1382445/14062855/80c9f808-f36b-11e5-805c-1ee9bf65fc8e.png)

For me I was able to get the bug when a Spinner follows an `H1` with no classes, but only for the first instance on the page. The second one with the same markup appears normal. I've inspected the `.ms-Spinner-circle` elements and they're always 3.58px by 3.58px in every spinner. This appears to be some sort of browser rounding bug.

@wdo3650, thoughts?

StfBauer:
Just stumbled upon this issue.
I would recommend to remove the javascript completely because it is not required to create such simple spinner and might cause more problems like this.

Instead use CSS3 for the rotation and propably a fallback as animated gif for older browsers if needed.

The spinner icon itself might be useful to be part of the icon font or will be served as SVG background image.

In general javascript should be avoided for such things because it can cause exceptions in browser and probably stop other more relevant scripts from working.
With CSS you don't have such issue and it simply work or doesn't work but in any case it won't break other code.

wdo3650:
[...]

PR #464, opened 2016-03-28
Tweaked the size of the dots in the small size to prevent strange ren…

wdo3650:
Fixes issue #462. Tweaked the size of the dots in the small size to prevent strange rendering happenings with h1's.


mikewheaton:
I've forked the repro for the previous issue and made your changes to it here: http://plnkr.co/edit/Cc40wA?p=preview

All looks good now. Nice work!

Jahnp:
Looks great, thanks for the fix! Approved.

Issue #442, opened 2016-03-18
Use font-weight to determine the web font to use


StfBauer:
Hi,

just spot that the [@fontface](https://github.com/OfficeDev/Office-UI-Fabric/blob/master/src/sass/_Fabric.Typography.Fonts.scss) definition is incorrect. All font declarations use the same font-weight but different font names. It is working but has a lot of negative side effects.

To be more specific the current font definition looks like this.
```
@font-face {
    font-family: "#{$ms-light} #{$ms-font-language}";
    src: local("Segoe UI Light"),
         url('#{$ms-font-directory}/#{$ms-font-path}/SegoeUI-Light.woff2') format('woff2'),
         url('#{$ms-font-directory}/#{$ms-font-path}/SegoeUI-Light.woff') format('woff'),
         url('#{$ms-font-directory}/#{$ms-font-path}/SegoeUI-Light.ttf') format('truetype');
    font-weight: normal;
    font-style: normal;
  }
```

instead the font face definition should look like.

```
@font-face {
    font-family: "#{$ms-font} #{$ms-font-language}";
    src: local("Segoe UI"),
         url('#{$ms-font-directory}/#{$ms-font-path}/SegoeUI-Light.woff2') format('woff2'),
         url('#{$ms-font-directory}/#{$ms-font-path}/SegoeUI-Light.woff') format('woff'),
         url('#{$ms-font-directory}/#{$ms-font-path}/SegoeUI-Light.ttf') format('truetype');
    font-weight: 300;
    font-style: normal;
  }
```

The $ms-light is not needed but instead I would recommend to use $ms-font. In addition for the light font the font-weight should be set to 300.

The following values are defined for the default font weights:
**100**    Thin (Hairline)
**200**    Extra Light (Ultra Light)
**300**    Light
**400**    Normal
**500**    Medium
**600**    Semi Bold (Demi Bold)
**700**    Bold
**800**    Extra Bold (Ultra Bold)
**900**    Black (Heavy)

This makes year css more cleaner and gives the browser the possibility to choose the correct font file. It also make the CSS file a little bit smaller.

mikewheaton:
Excellent points, Stefan. This could also simplify localization. Right now we override the `font-family` of every class that uses that uses that font weight. This is repeated for every language. For example:

```
*[lang='vi-VN'] .ms-font-mi,
*[lang='vi-VN'] .ms-fontWeight-semibold,
[...]

PR #410, opened 2016-03-04
Adding sample page for Language Overrides

notandrew:

PR #402, opened 2016-03-01
Shifting language overrides to opt-in by default.

notandrew:

Issue #382, opened 2016-02-24
Localization: Set font-family with ms-Fabric wrapper


kennywangjin:
when I apply "lang='cn-ZHS'" to <html />, the whole page font family keeps segoe ui, while it should be Microsoft yahei.
Is it a bug or have I missed some steps? 
(I am trying as office ui fabric website shows.)
Thanks in advance.

mikewheaton:
Could you take a look at this, @notandrew? Also, we should get a test/sample page set up so that we can quickly verify that the other font stacks are working.

notandrew:
We have a typo.  We are using ch-ZHS instead of cn.  This should fix it for this release of Fabric.  I'll make a fix tonight so it point to cn-ZHS.  I just noticed other teams use zh-cn. I'll investigate which language codes we should be using.

mikewheaton:
Would be great if we could use the [standard HTML ISO Language Codes](http://www.w3schools.com/tags/ref_language_codes.asp).

notandrew:
Microsoft has its own lcid codes: https://msdn.microsoft.com/en-us/library/cc233982.aspx

mikewheaton:
Interesting. Those are for Microsoft Windows though, do they make sense for a web product? We only use a tiny subset of that, and what we have maps more closely to the ISO 639-1 codes that the W3C recommends.

mikewheaton:
Also, he said that he's using "cn-ZHS", exactly as we have it in our docs. Is that not what's in Fabric right now?

notandrew:
Right, that's the typo, we have ch-ZHS in Fabric.

mikewheaton:
Ah okay. @kennywangjin, could you try changing the `lang` attribute to "ch-ZHS" and confirm that it works for you?

notandrew:
OneDrive uses the Microsoft lcid codes that I referenced above, but those match language/region instead of just language, which we're probably more interested in...

kennywangjin:
@mikewheaton I was using "ch-ZHS", just as the fabric web's demo. Not working.

mikewheaton:
Okay, I guess both cn-ZHS and ch-ZHS are not working for Kenny. We'll have to look into this and see if we can reproduce it.

notandrew:
@kennywangjin Are you applying our font classes to the containers? Here's what the Fabric CSS looks like -- so any of these font classes should be applied:
*[lang='ch-ZHS'] .ms-font-mi,
[...]

PR #486, opened 2016-04-11
Using font-weight to decide Segoe fonts, updating lang codes, opt-in to lang overrides

notandrew:
* Moving language overrides to opt-in instead of opt-out.
* Using ISO-639-3 language codes (https://en.wikipedia.org/wiki/ISO_639-3) 
* Updated font face declarations to all use one font name (Segoe UI WestEuropean) so that font-weight determines which Segoe variant (Semilight, Light, Regular, Semibold) gets displayed.
* Adding sample page for displaying override

Fixing:
#442 
#410 
#402 
#382


src/sass/Fabric.Typography.Output.scss
```diff
@@ -198,7 +198,7 @@
 
 .ms-fontColor-neutralSecondary, 
 .ms-fontColor-neutralSecondary--hover:hover {
- @include ms-fontColor-neutralSecondary;
+  @include ms-fontColor-neutralSecondary;
```
Jahnp: This looks like an unnecessary indentation. Please revert.

src/sass/Fabric.scss
```diff
@@ -19,7 +19,6 @@
 @import './Fabric.Typography.Variables';
 @import './Fabric.Typography.Output';
 @import './Fabric.Typography.Fonts.Output';
-@import './Fabric.Typography.Language.Overrides.Output';
```
Jahnp: Woohoo!

src/sass/_Fabric.Typography.Fonts.scss
```diff
@@ -31,48 +28,48 @@ $ms-font-path-westeuropean: "SegoeUI-WestEuropean";
...
+@mixin SegoeUISemilight($ms-font-language, $ms-font-path) {
   @font-face {
-    font-family: "#{$ms-regular} #{$ms-font-language}";
-    src: local("Segoe UI"),
-         url('#{$ms-font-directory}/#{$ms-font-path}/SegoeUI-Regular.woff2') format('woff2'),
-         url('#{$ms-font-directory}/#{$ms-font-path}/SegoeUI-Regular.woff') format('woff'),
-         url('#{$ms-font-directory}/#{$ms-font-path}/SegoeUI-Regular.ttf') format('truetype');
-    font-weight: normal;
+    font-family: "#{$ms-font-name} #{$ms-font-language}";
+     src: local("Segoe UI Semilight"),
```
Jahnp: This looks indented one space too far.

src/sass/_Fabric.Typography.Language.Overrides.scss
```diff
@@ -5,7 +5,6 @@
 // --------------------------------------------------
 // Language override definitions
 
-
```
Jahnp: Please leave this space.

src/sass/_Fabric.Typography.Variables.scss
```diff
@@ -5,11 +5,18 @@
 // --------------------------------------------------
 // Fabric Core Typography variables
 
+$ms-font-system-base: 'Segoe UI', -apple-system, BlinkMacSystemFont, 'Roboto', 'Helvetica Neue', sans-serif;
+$ms-font-family:  'Segoe UI WestEuropean', $ms-font-system-base;
```
Jahnp: Super small, but it feels like this should be more descriptive, e.g. `$ms-font-family-base` or something like that.

src/sass/Fabric.Typography.Output.scss
notandrew: It was only one space instead of two.  This fixes it to be consistent with the rest of the include rules.

Jahnp: Oh, I see--good catch!

mikewheaton:
Approved! Changes to `font-weight` pull in the appropriate font, without having to change the `font-family` name. This will allow us to finally localize the components' styles by setting a base font and then simply using `font-weight`.

mikewheaton:
We can go ahead and merge this without a second reviewer if you're ready to merge it. The sooner it's in the better – our next step will be removing all of the `$ms-fontWeight-semibold` type references within the components and using the `font-weight` property instead. 😄

PR #489, opened 2016-04-12
Convert Fabric Components to TypeScript

dammittom:
# TypeScript conversion

This pull request adds TypeScript support to our build tools, and converts the component source code to TypeScript. This is setting the framework for general component code cleanup in the future, with the primary objective to remove jQuery and more clearly separate operational and presentational JavaScript.

# Build Notes

This moves all of the JavaScript related gulp tasks into the ComponentJS.js file. Also, the samples html files for each component now includes the concatenated fabric .js file, instead of a standalone .js file with just the single component.

# Usage Notes

All of the code for the Fabric components are now under the Fabric namespace. The ones that are using jQuery are no longer implemented as jQuery plugins. The next step on our roadmap is to remove jQuery completely, which may be done in increments, so it was our goal to make this breaking change for all components all at once.

For example, any Fabric Dialog components in the DOM would be initialized this way now:
```js
var elements = document.querySelectorAll('.ms-Dialog');
var i = elements.length;
var component;
while(i--) {
  component = new fabric['Dialog'](elements[i]);
}
```


src/templates/componentSampleTemplate.html
```diff
@@ -11,22 +11,11 @@
   <link rel="stylesheet" href="<%= componentName %>.css">
 
   <script type="text/javascript" src="https://code.jquery.com/jquery-2.1.4.min.js"></script>
- 
+  <script type="text/javascript" src="../../../../js/fabric.js"></script>
+  <script type="text/javascript" src="../../../../lib/pickadate.js"></script>
```
mikewheaton: This isn't needed by all of the components, but it's not hurting anything to be there. Especially if we're able to remove this plugin entirely at some point.

mikewheaton:
Looks good to me! Let's get this into our `beta/3.0.0` branch for closer review, and to merge with the Handlebars templates when that branch is ready.

PR #491, opened 2016-04-13
Fixed the placeholder to support tab navigation

johnmeilleur:
Fixed the placeholder so it is hidden when using tab to navigate between text fields.


mikewheaton:
Looks good to me. One more approver and we can merge it in.

wdo3650:
Nice work @johnmeilleur. This looks good to me. Approved.

PR #492, opened 2016-04-13
Searchbox Move SASS function

dammittom:
Move SASS function to fabric colors file, giving the result its own variable. This is to help with another project that is using fabric, but tokenizes the colors so it can be extended with themes.

### DIFF
--- a/src/components/SearchBox/SearchBox.scss
+++ b/src/components/SearchBox/SearchBox.scss
@@ -55,7 +55,7 @@
 
   &.hovering {
     border-color: $ms-color-themePrimary;
-    background-color: $ms-color-primaryTranslucent20;
+    background-color: $ms-color-themeLighter;
 
     & + .ms-SearchBox-label {
       color: $ms-color-black;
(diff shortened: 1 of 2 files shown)


mikewheaton:
Where is this color being used? We should only include colors that are used widely and recommended by our designers as part of the core palette. The `transparentize()` mixin could be used by any application without baking this particular color into Fabric, right?

dammittom:
This color $ms-color-primaryTranslucent20 is only being used one, in the SearchBox sass file, so it does not meet the criteria of being widely used.

However, the two other translucent colors in our general fabric colors file, $ms-color-blackTranslucent40 and $ms-color-whiteTranslucent40, are only being used by the Overlay component. So with respect to these colors, it does fit into an existing pattern.

mikewheaton:
I'd consider it more of an anti-pattern. :smirk: Let's talk about it more in person as I know you have some extra considerations here. But generally I don't think we should be including color variables in Fabric that aren't part of the official palette and documented.

dammittom:
Updated this change so that it now uses an existing color (ms-color-themeLighter), which was actually the original design spec for the SearchBox.

Jahnp:
Awesome. Looks great--approved!

PR #498, opened 2016-04-15
Remove unused localhost Express server.js file

Jahnp:
This file wasn't being used by the build (it was effectively replaced by `gulp/modules/Server.js`).


mikewheaton:
Nice catch. :+1:

notandrew:
Looks super!

PR #499, opened 2016-04-15
Add fixed variant to large size panel

philworthington:
'Fixed' modifier class added for a large fixed panel variant. Check 'Container Behavior' guidelines in notebook.


mikewheaton:
I'm sure this meets the spec, but the behavior of the panels across screen sizes is getting a little complex, no? We should see if we can simplify this.

Jahnp:
Agreed on Mike's point that the overall system for Panels is already pretty complicated, and this doesn't really help. But, it's a valid scenario, and if we want to pare down the system in the future, we can revisit this piece then. For now, consider this approved. Nice work!

PR #503, opened 2016-04-20
Fixed radio button for RTL.

wdo3650:
- RadioButton was off to the right and not centered in the circle
- The check in CheckBox is improved with this PR; however in RTL this could use an RTL mixin to change the left position of the check to be centered correctly.


mikewheaton:
Looks good, approved.

Jahnp:
Awesome--approved.

PR #515, opened 2016-05-03
Adding MessageBar component

notandrew:
Adding MessageBar component to Fabric.

### DIFF
--- a/src/components/MessageBar/MessageBar.html
+++ b/src/components/MessageBar/MessageBar.html
@@ -2,15 +2,12 @@
 
 <div class="ms-MessageBar ms-MessageBar--info">
   <div class="ms-MessageBar-content">
-
     <div class="ms-MessageBar-icon">
       <i class="ms-Icon"></i>
     </div>
-
     <div class="ms-MessageBar-text">
       Info - lorem ipsum dolor sit amet, a elit sem interdum consectetur adipiscing elit.<br>
       <a href="" class="ms-Link">Hyperlink string</a>
     </div>
-
   </div>
 </div>
--- a/src/components/MessageBar/MessageBar.html
+++ b/src/components/MessageBar/MessageBar.html
@@ -1,0 +1,16 @@
+<!-- Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE in the project root for license information. -->
+
+<div class="ms-MessageBar ms-MessageBar--info">
+  <div class="ms-MessageBar-content">
+
+    <div class="ms-MessageBar-icon">
+      <i class="ms-Icon"></i>
+    </div>
+
+    <div class="ms-MessageBar-text">
+      Info - lorem ipsum dolor sit amet, a elit sem interdum consectetur adipiscing elit.<br>
+      <a href="" class="ms-Link">Hyperlink string</a>
+    </div>
+
+  </div>
+</div>
--- a/src/components/MessageBar/MessageBar.html
+++ b/src/components/MessageBar/MessageBar.html
@@ -1,9 +1,9 @@
 <!-- Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE in the project root for license information. -->
 
-<div class="ms-MessageBar ms-MessageBar--info">
+<div class="ms-MessageBar">
   <div class="ms-MessageBar-content">
     <div class="ms-MessageBar-icon">
-      <i class="ms-Icon"></i>
+      <i class="ms-Icon ms-Icon--infoCircle"></i>
     </div>
     <div class="ms-MessageBar-text">
       Info - lorem ipsum dolor sit amet, a elit sem interdum consectetur adipiscing elit.<br>
--- a/src/components/MessageBar/MessageBar.json
+++ b/src/components/MessageBar/MessageBar.json
@@ -1,0 +1,40 @@
+{
+  "name": "MessageBar",
+  "notes": "",
+  "description:": "A component for displaying information, alerts, and errors to a user via a small message surface.",
+  "template": "MessageBar.html",
+  "wrapBranches": true,
+  "class": "ms-MessageBar",
+  "dependencies": [
+    "Link"
+  ],
+  "fileOrder": [
+    "MessageBar.html",
+    "MessageBar.Alert.html",
+    "MessageBar.Error.html",
+    "MessageBar.Success.html",
+    "MessageBar.Remove.html"
+  ],
+  "branches": [
+    {
+      "name": "Info",
+      "class": "ms-MessageBar--info"
+    },
+    {
+      "name": "Alert",
+      "class": "ms-MessageBar--alert"
+    },
+    {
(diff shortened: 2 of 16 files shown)


mikewheaton:
I've added in a stub Markdown page to document this component. Could you complete it? I also had to remove the empty lines in the template because it was causing issues with our sample page generator.

src/components/MessageBar/MessageBar.html
```diff
@@ -0,0 +1,13 @@
+<!-- Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE in the project root for license information. -->
+
+<div class="ms-MessageBar ms-MessageBar--info">
```
mikewheaton: For other components we don't use a modifier for the default variant.

src/components/MessageBar/MessageBar.scss
```diff
@@ -0,0 +1,131 @@
...
+
+.ms-MessageBar--info {
+  @include ms-bgColor-info;
+
+  .ms-MessageBar-icon {
+    @include ms-fontColor-neutralSecondaryAlt;
+  }
+
+  .ms-Icon:before {
+    content: '\e00a';  // ms-Icon--infoCircle
```
Jahnp: This should be moved into the HTML as `<i class="ms-Icon ms-Icon--infoCircle"></i>`

src/components/MessageBar/MessageBar.scss
```diff
@@ -0,0 +1,131 @@
...
+    font-size: 16px;
+  }
+
+  .ms-MessageBar-text {
+    @include ms-font-s;
+  }
+
+}
+
+.ms-MessageBar--info {
```
Jahnp: By convention, modifiers should be joined with their root class, e.g. `.ms-MessageBar.ms-MessageBar--info` and preceded by a helpful comment following this form:
```scss
//== Modifier: Informational content
//
.ms-MessageBar.ms-MessageBar--info {
  ...
}
```

src/components/MessageBar/MessageBar.scss
```diff
@@ -0,0 +1,131 @@
...
+
+.ms-MessageBar--success {
+  @include ms-bgColor-success;
+
+  .ms-MessageBar-icon {
+    @include ms-fontColor-green;
+  }
+
+  .ms-Icon:before {
+    content: '\e21f';  // ms-Icon--checkboxCheck
```
Jahnp: Same as above--this should be moved to the HTML

src/components/MessageBar/MessageBar.scss
```diff
@@ -0,0 +1,131 @@
...
+
+.ms-MessageBar--warning {
+  @include ms-bgColor-warning;
+
+  .ms-MessageBar-icon {
+    @include ms-fontColor-warning;
+  }
+
+  .ms-Icon:before {
+    content: '\e00a';  // ms-Icon--info
```
Jahnp: Same as above--this should be moved to the HTML

src/components/MessageBar/MessageBar.scss
```diff
@@ -0,0 +1,131 @@
...
+
+.ms-MessageBar--error {
+  @include ms-bgColor-error;
+
+  .ms-MessageBar-icon {
+    @include ms-fontColor-error;
+  }
+
+  .ms-Icon:before {
+    content: '\e00c';  // ms-Icon--xCircle
```
Jahnp: Same as above--this should be moved to the HTML

src/components/MessageBar/MessageBar.scss
```diff
@@ -0,0 +1,131 @@
...
+
+.ms-MessageBar--remove {
+  @include ms-bgColor-error;
+
+  .ms-MessageBar-icon {
+    @include ms-fontColor-error;
+  }
+
+  .ms-Icon:before {
+    content: '\e005';  // ms-Icon--minus
```
Jahnp: Same as above--this should be moved to the HTML

src/sass/Fabric.Color.Mixins.Output.scss
```diff
@@ -15,98 +15,98 @@
   @include ms-bgColor-themeDark;
 }
 
-.ms-bgColor-themeDarkAlt,
```
Jahnp: Thanks for cleaning this file up! I'll approve these changes for this pull request, but in the future, it'd be good to keep changes scoped to just what's in the current review. Preferably, there would be a second PR for the changes in this file, so that these changes can be tracked separately of the MessageBar work.

src/sass/_Fabric.Typography.scss
```diff
@@ -344,8 +344,12 @@
   color: $ms-color-success;
 }
 
-@mixin ms-fontColor-alert {
-  color: $ms-color-alert;
+@mixin ms-fontColor-warning {
```
Jahnp: This is a breaking change. For now, can you add `ms-fontColor-warning` and mark `ms-fontColor-alert` and its variables as deprecated without removing them?

Jahnp:
Looks great! I left small formatting feedback and a note about rendering icons in HTML rather than CSS, but the biggest concern is around removing a variable/class that should probably be deprecated for the next release. Once those are addressed, consider this approved. Thanks!

src/sass/_Fabric.Typography.scss
mikewheaton: The documentation will also need to be updated to include the 'warning' and 'severeWarning' colors.

src/components/MessageBar/MessageBar.html
notandrew: fixed

src/components/MessageBar/MessageBar.scss
notandrew: Normally, that's what I'd do, but the nice thing about the way I built it is that we can switch the MessageBar type based on the main class.  Change ms-MessageBar--info to ms-MessageBar--error and you get the icon, color, and background switching automatically.  

I can go either way, but what do you think?

philworthington:
The scss has been cleaned up to include modifier comments, correct ordering, and icons were moved out of styling with correct classes added to template markup for each variant. Alert color added back with deprecation note in applicable files.

Jahnp:
Awesome--thanks for picking this up, @philworthington! Looks great. Looks like there might still be some merge conflicts to resolve--otherwise this looks ready to go. Consider it approved.

src/components/MessageBar/MessageBar.json
```diff
@@ -0,0 +1,41 @@
...
+    "MessageBar.Warning.html",
+    "MessageBar.SevereWarning.html",
+    "MessageBar.Error.html",
+    "MessageBar.Success.html",
+    "MessageBar.Remove.html"
+  ],
+  "branches": [
+    {
+      "name": "Info",
+      "class": "ms-MessageBar--info"
```
mikewheaton: What is this?

mikewheaton:
Approved

PR #518, opened 2016-05-03
Remove font-size-m-plus and font-size-s-plus

### LINKED ISSUE #517: Remove ms-font-m-plus and ms-font-s-plus
There are two sizes in our type ramp, `ms-font-m-plus` and `ms-font-s-plus`, that were intended for special use cases and not as part of the general ramp. To simplify the ramp and reduce confusion we're going to remove these sizes. This should be saved for the 3.0.0 release as it is a breaking, however minor.

mikewheaton:
Fixes #517 

A PR has been opened on the documentation site. This should not be merged until that PR is approved and merged.

### DIFF
--- a/src/components/ChoiceField/ChoiceField.scss
+++ b/src/components/ChoiceField/ChoiceField.scss
@@ -130,7 +130,7 @@
 			content: '\e041';
 			background-color: transparent;
 			border-radius: 0;
-			font-size: $ms-font-size-s-plus;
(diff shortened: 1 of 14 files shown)


mikewheaton:
Now that `beta/3.0.0` has been merged into `master`, I'm going to close this PR for now. We should have an answer from design soon on whether or not these classes are being removed. If they are, I'll open a new PR directly into `master`.

Issue #332, opened 2016-02-08
Border color classes don't set a border


mikewheaton:
Our border color classes (e.g. `.ms-borderColor-themePrimary`) set the `border-color` CSS property, but do not provide a `border-width` or `border-style`. This leads to confusion when developers apply a border color and see nothing.

At a minimum we should document this more clearly, and provide an example of a class they can apply that will create a border. To help further, we could bake in a 1px solid border for all of the border color classes (this could be easily overridden) or creating a `.ms-u-border` type utility to be used in conjunction with the color classes.

Thoughts?

notandrew:
It's weird either way.  Having two border classes (one to set up the size and style and one to set color) seems like overkill but otherwise it would be easy to assume that you're going to get a border by applying the color class. 

Baking in the 1px solid border for color classes seems okay since that's what the design language calls for in most places...

notandrew:
On second-thought, it's a color class that should probably not impact border/style. Hadn't thought that one through.

mikewheaton:
I'm leaning toward adding the border width and style to the color classes. This is similar to the base typography classes, where we set multiple properties based on common usage and then let people override these as necessary. Our designs almost always call for a one pixel solid border.

mikewheaton:
We discussed this today and decided to create a new utility class, `.ms-u-borderBase` that will be something like this:

```
.ms-u-borderBase {
  border: 1px solid $color-neutralPrimary;
}
```

To get a colored border, you would then apply both `.ms-u-borderBase` and `.ms-u-borderColor-[color]`. this follows the same pattern of a base class with overrides that we use for fonts.

mikewheaton:
Fixed by #524. 😄

PR #524, opened 2016-05-06
Added a new utility class to set border base width

atneik:
Added .ms-u-borderBase to set default base width. One would then apply
both .ms-u-borderBase and .ms-borderColor-[color] to see a colored
border.

https://github.com/OfficeDev/Office-UI-Fabric/issues/332

### DIFF
--- a/src/sass/Fabric.Utilities.Output.scss
+++ b/src/sass/Fabric.Utilities.Output.scss
@@ -12,6 +12,10 @@
   @include ms-u-borderBox;
 }
 
+.ms-u-borderBase {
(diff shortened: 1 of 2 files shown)


mikewheaton:
Looks good to me. We'll just need some documentation updates before merging this in.

Jahnp:
Looks great! Approved. Echoing Mike's feedback around doc updates, though.

atneik:
Added documentation updates to the issue: https://github.com/OfficeDev/Office-UI-Fabric/issues/536

PR #525, opened 2016-05-06
Remove condition preventing is-active CommandBarSearch

maxali:
There is if($(window).width() < 640) condition to add is-active class to .ms-CommandBarSearch. 

But `ms-CommandBarSearch` is already closed by `@media screen` at [CommandBar.scss#L230](https://github.com/OfficeDev/Office-UI-Fabric/blob/master/src/components/CommandBar/CommandBar.scss#L230).

For `screens < /*widthLgCollapsed*/ (1023px) > 640px` will never trigger the `.click` event.

### DIFF
--- a/src/components/CommandBar/Jquery.CommandBar.js
+++ b/src/components/CommandBar/Jquery.CommandBar.js
@@ -89,13 +89,10 @@
       processOverflow(overFlowCommands, $OverflowCommand, $OverflowMenu);
 
       // Set Search Behavior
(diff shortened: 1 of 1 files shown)


mikewheaton:
It looks like this bug was fixed separately in PR #615. Thank you for your contribution here – our apologies for not getting it merged in before we started all of the work on Fabric 3.0 that only recently made it into the master branch.

PR #526, opened 2016-05-06
Add presence icons to handlebars persona

philworthington:
Presence icon scss, markup, json added to include applicable skype icons for persona.

NOTE: Base64 icon font added. Should be removed and cdn icon font uncommented when icon font is available.


battletoilet:
Its gooodz im pulling it in.

PR #529, opened 2016-05-07
Fix simple Sass linting issues

mikewheaton:
I've gone through and fixed most (over 400!) of the linting issues, which were primarily caused by empty spaces or missing new lines between selectors. I made some adjustments to the Sass linting settings which we'll want to review, either in this PR or before release. There are a lot of rules that are a matter of style so I leaned towards matching whatever I saw happening most often in our code.

There are two remaining warnings that we'll want to discuss:
1. The use of `!important`. We have this in Pivot and Toggle, and should see if we can remove it.
2. Only using `@extend` with [placeholder selectors](http://thesassway.com/intermediate/understanding-placeholder-selectors). This is a new Sass feature I'm unfamiliar with. We only have this error for one line of PersonaCard.

Note that the build will still fail due to TypeScript linting errors, which we can handle in another PR.

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -47,3 +47,10 @@ rules:
       properties:
         - margin
         - content
+
+	# Allow empty lines between blocks and single line rulesets
+	# https://github.com/sasstools/sass-lint/blob/master/docs/rules/empty-line-between-blocks.md
+	empty-line-between-blocks:
+    - 1
+    -
+      allow-single-line-rulesets: true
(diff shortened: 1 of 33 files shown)


battletoilet:
This looks great, thanks for fixing these man!

ericthompson:
This is awesome - good work, @mikewheaton!

mikewheaton:
No problem, it makes for a fun couple of hours. 😉 

All of our Travis builds will fail until we get these `!important` flags removed, so we'll want to manually review the build log before merging any new PRs.

PR #530, opened 2016-05-07
Fix TypeScript linting issues

mikewheaton:
Went through and fixed all of the TypeScript linting errors and warnings, most of which were related to whitespace or not specifically declaring variables as public or private. I also commented out some unused functions (these should be reviewed and removed if they won't be used) and removed unused variables.

The only change I've made to the linting rules is to allow 'inferable types', such as `boolean = true`. It's not strictly necessary to provide a type for these but VSCode shows an error without a type. I also don't see the harm in being very explicit about what type is desired.

### DIFF
--- a/src/components/AlignmentHost/AlignmentHost.ts
+++ b/src/components/AlignmentHost/AlignmentHost.ts
@@ -4,7 +4,8 @@
  * ContextualHost
  *
  * Hosts contextual menus and callouts
- * NOTE: Position bottom only works if html is set to max-height 100%, overflow hidden and body is set to overflow scroll, body is set to height 100%
+ * NOTE: Position bottom only works if html is set to max-height 100%, overflow
+ * hidden and body is set to overflow scroll, body is set to height 100%
  *
  */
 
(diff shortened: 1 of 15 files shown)


src/components/AlignmentHost/AlignmentHost.ts
```diff
@@ -69,30 +68,30 @@ namespace fabric {
...
-        case "top":
-          break;
-        case "bottom":
-          break;
-        default:
-        this._setPosition();
-      }
-    }
+
+    // private _getAlignmentSize() {
```
Jahnp: I'm assuming this will be either revised or removed at some point--we shouldn't leave commented-out blocks of code in PRs if possible.

src/components/AlignmentHost/AlignmentHost.ts
```diff
@@ -269,36 +268,36 @@ namespace fabric {
...
-        + (parseInt(_alignmentStyles.marginTop, 10)
-        + parseInt(_alignmentStyles.marginBottom, 10));
-     
-      this._alignmentClone.classList.remove(ALIGNMENT_STATE_POSITIONED);
-      this._alignmentClone.classList.remove(CONTEXT_STATE_CLASS);
-      this._teWidth = this._targetElement.getBoundingClientRect().width;
-      this._teHeight = this._targetElement.getBoundingClientRect().height;
-    }
+
+    // private _savealignmentSize(): void {
```
Jahnp: Same feedback as above--is there a reason this is commented out? The code looks like it's been cleaned up.

battletoilet: It will be when we continue working on the JS again.

battletoilet:
Approved

Jahnp:
Looks great! Approved.

mikewheaton:
Perfect, I'll get this merged in. The commented-out functions should definitely be removed in the future if they're not used.

PR #535, opened 2016-05-10
Added animation library to Fabric.

wdo3650:


mikewheaton:
Try merging `beta/3.0.0` into this branch and see what linting errors remain.

mikewheaton:
Approved

Jahnp:
This is *really*, really cool, @wdo3650! Super excited for these additions. Approved. 

At some point--I don't want to block on this right now--one thing I'd love us to do is think about how we document these additions. I'm thinking on the documentation site, in the .md files under ghdocs, and in-code (e.g. JSDoc). We should include guidance for where, how, and when to use these methods, ideally with demonstrational examples.

PR #541, opened 2016-05-12
Merge latest master into beta branch

mikewheaton:
To follow up on @battletoilet's PR that we just merged, here are some more changes that have been made in master that we want to pull into the beta branch. We're not planning any more updates to master so this will likely be the last merge like this.


atneik:
Looks good!

PR #543, opened 2016-05-13
Fix Sass linting warnings by removing !important and @extend

mikewheaton:
This PR fixes the remaining Sass linting warnings by removing `!important` from Pivot and Toggle. I've tested these (screenshots below) and they display fine without the extra specificity.

Also removed `@extend` from PersonaCard, as this was resulting in a linting warning and we get the equivalent output with a simple selector.

screen shot 2016-05-13 at 9 36 28 am
screen shot 2016-05-13 at 9 39 26 am

### DIFF
--- a/src/components/PersonaCard/PersonaCard.scss
+++ b/src/components/PersonaCard/PersonaCard.scss
@@ -36,7 +36,8 @@
   height: 48px;
 }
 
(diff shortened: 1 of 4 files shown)


mikewheaton:
Aniket says approved! Merging it in so we don't get merge conflicts

PR #545, opened 2016-05-14
ChoiceField, Toggle: Removed CSS sibliing selectors

atneik:
Using TS instead of CSS sibling selectors.
Other markdown and handlebar fixes.


battletoilet:
Approved

PR #550, opened 2016-05-17
Generate component examples with Handlebars

mikewheaton:
This PR adds examples for components to the documentation page, generated using the Handlebars templates. It covers about a third of the components (additional PRs will come in from @battletoilet and @atneik) and there are still many bugs, but it's a start.


battletoilet:
Approved

PR #553, opened 2016-05-17
Update PullApprove approvers list

mikewheaton:
Updated the PullApprove configuration. I'm not 100% sure this is necessary in every branch, as I think PullApprove just looks at master. But we needed to do a test pull request anyway.


Jahnp:
Approved.

mikewheaton:
Approved by me as well

mikewheaton:
Woohoo! Looks like it's working 😄

PR #556, opened 2016-05-18
Template docs

atneik:


src/components/ChoiceFieldGroup/ChoiceFieldGroup.hbs
```diff
@@ -4,6 +4,6 @@
     <label class="ms-Label {{#if props.required}} is-required {{/if}}">{{props.title}}</label>
   </div>
   {{#each props.fields}}
-    {{renderPartial component props}}
+    {{> (lookup . 'component') props=props}}
```
mikewheaton: We should be able to change these to use `renderPartial` after the merge.

mikewheaton:
Approved

src/components/ChoiceFieldGroup/ChoiceFieldGroup.hbs
atneik: Sounds good!

PR #565, opened 2016-05-18
Update how text properties are set for components

mikewheaton:
Now that we're using `font-weight` to control the font weight, rather than `font-family`, the Segoe web font was not loading correctly in any of the components.

This PR adds the base font family to every component's root element (e.g. `.ms-Button`) so that all child elements inherit the correct font. It then sets the `font-weight` wherever `font-family` was being used previously.

I've also replaced all of the `@include` calls which were setting `font-family`, `font-weight`, `font-size`, and `color` all at once. This can be convenient but it often leads to output with duplicate properties such as:

```
.ms-Component {
  font-family: ...
  font-weight: 100;
  font-size: 16px;
  color: #666666;
  font-weight: 300;
  font-size: 14px;
}
```

It also converts all tabs to spaces.

### DIFF
--- a/src/components/PanelHost/PanelHost.scss
+++ b/src/components/PanelHost/PanelHost.scss
@@ -2,106 +2,95 @@ $Panel-width-lightDismiss: 272px;
 
 // The panel covers the entire screen.
 .ms-PanelHost {
-	bottom: 0;
-	left: 0;
-	position: absolute;
-	right: 0;
-	top: 0;
-	pointer-events: none;
-	z-index: $ms-zIndex-front;
-	font-family: $ms-font-family-base;
+  bottom: 0;
+  left: 0;
+  position: absolute;
+  right: 0;
+  top: 0;
+  pointer-events: none;
+  z-index: $ms-zIndex-front;
+  font-family: $ms-font-family-base;
 }
-
-// .ms-PanelHost-inner {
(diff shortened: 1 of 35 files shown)


battletoilet:
Approved

src/components/PanelHost/PanelHost.scss
```diff
@@ -2,105 +2,106 @@ $Panel-width-lightDismiss: 272px;
...
-  pointer-events: none;
-  z-index: $ms-zIndex-front;
+	bottom: 0;
+	left: 0;
+	position: absolute;
+	right: 0;
+	top: 0;
+	pointer-events: none;
+	z-index: $ms-zIndex-front;
+	font-family: $ms-font-family-base;
```
philworthington: Convert to spaces

src/components/Table/Table.scss
```diff
@@ -6,130 +6,131 @@
 // Data table styles
 
 .ms-Table {
-  display: table;
-  width: 100%;
-  border-collapse: collapse;
+	display: table;
+	width: 100%;
+	border-collapse: collapse;
+	font-family: $ms-font-family-base;
```
philworthington: Convert to spaces

philworthington:
Approved

PR #566, opened 2016-05-18
Fix 201402: 3.0.0: Toggle: Tabbing/selection doesn't work

philworthington:
Tabbing and selection with spacebar added to toggle.
Gulpfile updated so that all files are built on change (scss).


mikewheaton:
Approved 👍

battletoilet:
Approved

PR #570, opened 2016-05-19
ChoiceField, ChoiceFieldGroup updates

atneik:
Following aria guidelines. ChoiceField send custom events to
choiceFieldGroup for radio selection.


src/components/ChoiceField/ChoiceField.hbs
```diff
@@ -1,10 +1,12 @@
...
-<div class="ms-ChoiceField">
-  <input 
-    id="{{props.id}}" 
-    class="ms-ChoiceField-input"
-    type="{{props.type}}"
-    name="{{props.name}}" {{#if props.disabled}} disabled {{/if}} {{#if props.checked}} checked {{/if}} />
-  <label for="{{props.id}}" class="ms-ChoiceField-field"><span class="ms-Label">{{props.label}}</span></label>
+<div class="ms-ChoiceField"> 
+  <li role="{{props.type}}"
+      id="{{props.id}}"
```
mikewheaton: Is an `id` necessary? It seems to work fine without it.

src/components/ChoiceFieldGroup/ChoiceFieldGroup.hbs
```diff
@@ -1,9 +1,11 @@
 <!-- Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE in the project root for license information. -->
-<div class="ms-ChoiceFieldGroup">
+<div class="ms-ChoiceFieldGroup" id="{{props.id}}">
   <div class="ms-ChoiceFieldGroup-title">
     <label class="ms-Label {{#if props.required}} is-required {{/if}}">{{props.title}}</label>
   </div>
+  <ul class="ms-ChoiceFieldGroup-list">
```
mikewheaton: Add `role=radiogroup` to indicate that this element contains all of the `radio` elements.

mikewheaton:
This is great! So nice to remove the hidden inputs and CSS trickery.

Let's talk about renaming these components to 'CheckBox' and either 'ChoiceGroup' or 'RadioButtonGroup'. That would make the difference and usage clearer, and match another project where we're doing the same. 😄

src/components/Dialog/Dialog.json
```diff
@@ -1,5 +1,248 @@
...
+    "Dialog.Multiline.html",
+    "Dialog.Blocking.html"
+  ],
+  "dependencies": [
+    "Button",
+    "CheckBox",
+    "Overlay",
+    "ContextualHost"
+  ],
+  "props": {
```
mikewheaton: We don't need props, dependencies, fileOrder, etc. in the JSON anymore

atneik: Not sure what is used by the build. But, props seem safe to remove.

PR #574, opened 2016-05-19
Dialog fixes

mikewheaton:
This PR updates the styles for Dialog to match the redlines. It also adds some basic demonstration JS to the example page. This still needs some work before it's ready for release.


battletoilet:
Approved

battletoilet:
Although we will want to add a bit more functionality to the dialog but this is good.

PR #578, opened 2016-05-20
Add keyboard accessibility to OrgChart

mikewheaton:
Removes tabindex from Persona (only some variants are actionable) and adds tabindex to the OrgChart.


wdo3650:
Approved

PR #582, opened 2016-05-20
ProgressIndicator demo

mikewheaton:
This PR adds a basic demo for the ProgressIndicator, where it fills up after 2 seconds. Also added animation so the progress doesn't jump.

image


battletoilet:
Approved

PR #583, opened 2016-05-20
MessageBanner demonstration and focus states

mikewheaton:
Added a simple button to open the MessageBanner, and removed our `outline: none` override that was preventing the close button for showing a focus state. We'll want to style this better in the future but at least it's accessible for now.

image


battletoilet:
Approved

PR #590, opened 2016-05-21
Wired up dialogs with action, blocking, overlays

atneik:


mikewheaton:
Approved! It's working perfectly, thanks for fixing this up so well 😄

src/documentation/pages/Dialog/examples/DialogExampleDefaultJS.hbs
```diff
@@ -0,0 +1,52 @@
...
+    labelText += counter + " option(s) selected. " + this.innerText.trim() + " clicked" ;
+    label.innerText = labelText;
+  }
+  
+  function openDialog(dialog) {
+    // Open the dialog
+    dialogComponent.open();
+
+    // Center the dialog
+    dialog.style.top = (window.innerHeight - dialog.offsetHeight) / 2 + 'px';
```
battletoilet: Instead of setting an exact pixel value for centering the Dialog, let's make this happen in css. You can add this code below in somewhere on the main dialog and you can basically remove the positioning css.

```
top: 50%;
left: 50%;
transform: translate(-50%, -50%);
```

mikewheaton: We've had problems in the past where transform put the element on a half pixel and caused blurring. I don't recall the specifics of what browser that was in. However, based on this [StackOverflow thread](http://stackoverflow.com/questions/18257132/is-it-possible-to-snap-to-pixel-after-a-css-translate) it seems that it should work just fine as long as we avoid `translate3d`:

image

So let's give this a shot in CSS. 😄

battletoilet:
Approved

PR #591, opened 2016-05-23
Wdo3650/personacard fixes

wdo3650:
Added shadow and conditional removal if hosted by ContextualHost
Added tabbing

### DIFF
--- a/src/components/PersonaCard/PersonaCard.scss
+++ b/src/components/PersonaCard/PersonaCard.scss
@@ -48,6 +48,7 @@
   line-height: 48px;
   padding: 0 10px;
   color: $ms-color-neutralSecondary;
+  outline: transparent;
   position: relative;
   @include ms-u-borderBox;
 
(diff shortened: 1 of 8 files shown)


src/components/PersonaCard/PersonaCard.scss
```diff
@@ -181,29 +182,17 @@
...
-
 @media (min-width: $ms-screen-md-min) {
   // Undock the persona card and give it a shadow.
   .ms-PersonaCard {
+    @include drop-shadow;
     max-width: 360px;
     position: relative;
   }
+
+  .ms-ContextualHost {
```
mikewheaton: This is a great solution for now, but we should discuss whether ContextualHost should be applying to shadow or just positioning the component.

mikewheaton:
Approved 👍

src/components/PersonaCard/PersonaCard.scss
battletoilet: Yeah we should eventually just have ContextualHost do the positioning.

PR #597, opened 2016-05-24
Import MessageBar into the main components' Sass file

### LINKED ISSUE #596: MessageBar specific classes are missing in fabric.components.css 2.6.0
Building the current project results in css files without the messagebar component.

mikewheaton:
Fixes #596 

Will be included in a 2.6.1 patch release.


battletoilet:
Approved

mikewheaton:
I'm going to break all the rules and go ahead with a single reviewer, given how tiny this change is and that I want to get 2.6.1 out right away.

PR #599, opened 2016-05-24
Removed unnecessary chevron and fixed margins for overflow ellipsis.

wdo3650:
Removed unnecessary chevron at overflow and fixed margins.


mikewheaton:
Approved

Issue #552, opened 2016-05-17
Dropdown: Not working in mobile view


techvenky:
Fabric UI dropdown component is not working in mobile view or smaller width. Please find the attached image.

![image](https://cloud.githubusercontent.com/assets/1742973/15338183/bd73e5fc-1c4c-11e6-9ca4-fdf0c41f248c.png)

mikewheaton:
Thanks for reporting this, we'll have a look.

lixonic:
this is happening below 480px . Opened a PR . please review
 
![issue1](https://cloud.githubusercontent.com/assets/934220/15550070/56318318-22cd-11e6-92f7-81f9659f0f42.gif)

mikewheaton:
Fixed by PR #615

andrewconnell:
While fixed in current versions of fabric, any thoughts to going back and applying the CSS fix to the 2.6.2 release? Many folks didn't update to the latest MDL2 implementation of the library yet.

mikewheaton:
@andrewconnell: We should be able to do that. Would you mind opening an issue in [Fabric JS](https://github.com/OfficeDev/office-ui-fabric-js) for this?

PR #602, opened 2016-05-25
Fix Dropdown in screens below 480px

lixonic:
related to #552


mikewheaton:
We appreciate your contribution here, but it turns out this issue was actually resolved in our work preparing Fabric 3.0 which just merged in (PR #615). Sorry that we didn't update the issue so that you'd have known we were working on it.

PR #603, opened 2016-05-25
Wdo3650/beta/mdl2/breadcrumb

wdo3650:
Breadcrumb MDL2-ified

PR #610, opened 2016-05-27
Button & Toggle as mdl2

dammittom:
This updates the Button and Toggle components to their mdl2 design

### DIFF
--- a/src/components/Toggle/Toggle.scss
+++ b/src/components/Toggle/Toggle.scss
@@ -13,15 +13,17 @@
   // Slider pseudo element
   &:before {
     position: absolute;
-    top: 4px;
+    top: 3px;
     width: 10px;
     height: 10px;
-    transition: .1s ease;
     border-radius: 10px;
     content: '';
     #{$direction}: 4px;
-    background-color: $ms-color-neutralSecondaryAlt;
-    outline: 2px solid transparent;
+    background-color: $ms-color-neutralSecondary;
+    outline: 1px solid transparent;
+    transition-property: background, left;
+    transition-duration: 250ms;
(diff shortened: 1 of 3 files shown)


mikewheaton:
To get the build working you'll need the `DetailsList.hb` file from PR #603. You can grab it from there or wait until that PR merges, and then merge `beta/mdl2` back into yours. 😄

dammittom:
nice! Merged in the updated target branch

src/components/Toggle/Toggle.scss
```diff
@@ -180,11 +205,20 @@
...
-  width: 57px;
+  width: 45px;
   height: 20px;
   box-sizing: border-box;
-  border: 2.5px $ms-color-neutralTertiaryAlt solid;
+  border: 2px solid $ms-color-neutralTertiary;
+  border-radius: 20px;
   cursor: pointer;
+  transition-property: background, left;
+  transition-duration: 250ms;
```
Jahnp: Since it's so close, can you just use $ms-duration2?

src/components/Toggle/Toggle.scss
```diff
@@ -180,11 +205,20 @@
...
+  transition-timing-function: cubic-bezier(.4, 0, .23, 1);
```
Jahnp: Same here--I won't block approval on this, but I'm wondering if we should just use one of `$ms-ease1` or `$ms-ease2`.

Otherwise, we should probably either:
1. Capture this as a local variable for Toggle.scss (probably the safest for now), or
2. Incorporate this as a global animation

Jahnp:
Approved

PR #614, opened 2016-06-01
Update toggle MDL2 styling

philworthington:
Update toggle with correct border-color on hover and focus. Add border-color to transition-property.

### DIFF
--- a/src/components/Toggle/Toggle.scss
+++ b/src/components/Toggle/Toggle.scss
@@ -141,6 +141,7 @@
       &:focus,
       &:hover {
         background-color: $ms-color-themeDarkAlt;
+        border-color: $ms-color-themeDarkAlt;
       }
     }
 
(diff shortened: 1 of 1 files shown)


src/components/Toggle/Toggle.scss
```diff
@@ -211,9 +212,10 @@
   border: 2px solid $ms-color-neutralTertiary;
   border-radius: 20px;
   cursor: pointer;
-  transition-property: background, left;
+  transition-property: background, left, border-color;
   transition-duration: 250ms;
   transition-timing-function: cubic-bezier(.4, 0, .23, 1);
+  outline: 0;
```
Jahnp: Is this ok for high contrast mode? Are the border styles here enough?

Jahnp:
Approved

mikewheaton:
Approved 👍

src/components/Toggle/Toggle.scss
philworthington: The outline was going around the label area of the toggle which would not be the appropriate high contrast style. We might have to do another high contrast pass of all of the components after the mdl2 updates have been made.

PR #625, opened 2016-06-04
Remove ms-font-m-plus and ms-font-s-plus

beneficialmedia:
Figured I should finally start contributing. This fixes [Issue 517](https://github.com/OfficeDev/Office-UI-Fabric/issues/517)


mikewheaton:
We're actually having an internal debate right now about whether we're going to remove these sizes or not. I should have updated [the issue](https://github.com/OfficeDev/Office-UI-Fabric/issues/517) to say that. Let's keep the PR open for now, but unfortunately it may have to be closed if the decision is to leave these sizes in.

beneficialmedia:
I've an internal thread I'll forward to you on Monday. We should align with the Office 365 font-ramp.

mikewheaton:
After much discussion, it's been decided to keep these sizes in the type ramp.

PR #649, opened 2016-06-07
MDL2 update for checkbox and radiobutton

philworthington:
CheckBox and RadioButton updated to MDL2 styling. RadioButton no longer dependent on CheckBox in scss or ts implementation. Class names updated.

Linting gulp task fixed - task now finishes if there are warnings.


src/components/RadioButton/RadioButton.ts
```diff
@@ -9,30 +9,59 @@ namespace fabric {
    * Adds basic demonstration functionality to .ms-RadioButton components.
    *
    */
-  export class RadioButton extends CheckBox {
+  export class RadioButton {
```
atneik: Why are we removing RadioButton extending CheckBox? We will have to implement the same functionality in RadioButton that are already there in CheckBox, hence prevent code reuse/DRY. RadioButton is a special type of CheckBox.

mikewheaton: Phil can provide more detail, but I think the issue he was seeing was that there's a delay when JS turns a checkbox into a radio button. So it appears as a checkbox for a second and then turns into a radio button.

atneik: I see! Seems like we will have to have a base class, lets say ChoiceField, without any style from which both CheckBox and RadioButton could be extended. Keeping common functionality in base class, but style and other functionality in child classes.

philworthington: Yeah I like the idea of having a base class, ChoiceField. RadioButton extending CheckBox may be dry but requiring checkbox every time you want to use radiobutton seemed like an unnecessary coupling. ChoiceField could also hold common styles between the two as well, like how it was originally implemented. We should hold off from reviewing this, looks like there needs to be some exploration / extra work done here to implement ChoiceField.

mikewheaton:
We'll need a page to demonstrate the RadioButton component:

image

philworthington:
The checkmark not showing should be fixed. We are moving forward with the radio button as separate and not inheriting from check box as these are considered to be two separate components. This PR is ready to be reviewed.

mikewheaton:
Approved 🔨 

Turns out I just had to run `gulp nuke` before building, as it was holding onto the previously-built `fabric.css` rather than overwriting it.

src/components/CheckBox/CheckBox.scss
```diff
@@ -19,106 +22,122 @@
...
+  position: absolute;
+  box-sizing: border-box;
+  transition-property: background, border, border-color;
+  transition-duration: $ms-checkbox-transition-duration;
+  transition-timing-function: $ms-checkbox-transition-timing;
+}
+
+// The check mark icon
+.ms-CheckBox-field::after {
+  content: '\E73E';
```
Jahnp: We should be using the icon mixins here instead of hard-coded Unicode values, e.g. `@include ms-Icon--CheckMark`

Jahnp:
Approved 🌵 

I left one small comment about using mixins instead of hard-coded Unicode values, and it looks like there are a bunch of merge conflicts since some things have been checked in since this has opened. Once those are addressed feel free to merge this in--looks great!

src/components/CheckBox/CheckBox.scss
philworthington: Good call, this has been updated.

src/components/ContextualMenu/ContextualMenu.scss
```diff
@@ -142,7 +142,7 @@
       background-color: $ms-color-white;
 
       // Checkmark
-      &:after {
+      &::after {
         @include ms-Icon;
         color: $ms-color-neutralPrimary;
         content: '\e041';
```
mikewheaton: Use a mixin instead of this value.

src/components/Dropdown/Dropdown.scss
```diff
@@ -111,7 +111,7 @@
     padding: 0 16px;
   }
 
-  &:before {
+  &::before {
     content: none;
```
mikewheaton: Is `content: none` the same as `content: ''`? If so, we should use `none` as it's clearer what the intent is.

src/components/ContextualMenu/ContextualMenu.scss
philworthington: Getting this in next PR. It's going to be fantastic.

src/components/Dropdown/Dropdown.scss
philworthington: Scope creep. ✋

PR #652, opened 2016-06-08
Hero button spacing

### LINKED ISSUE #345: Hero button relies on space in HTML

mikewheaton:
Fixes #345.

Now that IE9 support is no longer a requirement, we can fix this pesky bug. The previous `display: inline-block` method of placing the icon next to the label was sensitive to spaces in the HTML. This now uses `display: flex` to position the elements. An added benefit is that we don't have to approximate vertical centering by using line height – we now have proper vertical centering that even looks correct in IE10.

Chrome
image

IE10
image

Thanks to @chrisahhh for the [original pull request](https://github.com/OfficeDev/Office-UI-Fabric/pull/434) and suggestions on how to fix this.

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -125,6 +125,7 @@
 //== Modifier: Hero button
 //
 .ms-Button.ms-Button--hero {
+  align-items: center;
   background-color: transparent;
   border: 0;
   display: flex;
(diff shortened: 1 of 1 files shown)


Jahnp:
Approved

philworthington:
Approved

PR #653, opened 2016-06-08
Use the disabled attribute for disabled TextFields

### LINKED ISSUE #500: Textfield: use disabled attribute, not is-disabled class
The textfield is not disabled in IE 10 when class .is-disabled is used.

This markup:
`<div class="ms-TextField is-disabled">
  <label class="ms-Label">Name</label>
  <input class="ms-TextField-field">
  <span class="ms-TextField-description">This should be your first and last name.</span>
</div>`

should render a textfield like that one (ie. latest Chrome)
image

in IE 10 it is still possible to click into the inputfield and insert stuff:
image

I think the source of the problem is, that the class .is-disabled is using " pointer-events: none;". It seems that this is not working in IE 10. In IE 10 it is required to set the input to disabled as well.

mikewheaton:
Fixes #500 

This PR adds a disabled property to the TextField, which both adds the `is-disabled` state class (to style the label) and adds a `disabled` attribute to the field itself. This ensures that the field is actually disabled across all browsers and doesn't just appear so.

image

### DIFF
--- a/src/components/TextField/TextField.hbs
+++ b/src/components/TextField/TextField.hbs
@@ -1,8 +1,8 @@
 <!-- Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE in the project root for license information. -->
 
-<div class="ms-TextField {{#if props.modifier}} ms-TextField--{{props.modifier}}{{/if}}">
+<div class="ms-TextField {{#if props.modifier}} ms-TextField--{{props.modifier}}{{/if}} {{#if props.disabled}}is-disabled{{/if}}">
   {{#if props.label}}<label class="ms-Label">{{props.label}}</label>{{/if}}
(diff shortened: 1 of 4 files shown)


Jahnp:
Should the SCSS be updated to add the appropriate styles if the disabled attribute is added?

mikewheaton:
Hmm, that's an interesting question. Adding the `disabled` attribute probably should style the field correctly, although that attribute alone won't style the corresponding label. So we would never recommend adding the attribute without the `is-disabled` state class being used as well.

Still, it's an easy mistake to make and a disabled text field should look the part, so I'll add that.

mikewheaton:
I've added styling for the `disabled` attribute on `ms-TextField-field`.

Jahnp:
Approved

Jahnp:
Thanks @mikewheaton!

battletoilet:
Approved

PR #658, opened 2016-06-10
Prevent text moving in pivot when font weight changes

### LINKED ISSUE #288: Prevent text moving in pivot when font weight changes
This is more a user experience issue than a bug. 

In the pivot component the selected element is bold and color highlighted. In the large pivot the element is defined only be the color.
The problem that I like to address is that the regular pivot elements start to jump around right after the selected element have been deselected. (Simply be clicking the bubbles for example). This issue is caused by changing the font-weight.
This flickering of the font-weight is very irritating because you change something on the UI without actually changing something.

mikewheaton:
Fixes #288

This PR uses the `title` attribute for the each pivot item to create a pseudo-element with the same text in bold. This element increases the width of the pivot, so that when the text changes to bold it doesn't suddenly change the pivot width.

See the [demonstration GIF](
http://cl.ly/0Y0W2T440l0V/Screen%20Recording%202016-06-10%20at%2012.37%20PM.gif).


Jahnp:
Approved

[Approved with PullApprove](https://pullapprove.com/OfficeDev/Office-UI-Fabric/pull-request/658/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=OfficeDev/Office-UI-Fabric)

battletoilet:
Looks super good mike

battletoilet:
Approved

PR #659, opened 2016-06-10
Use CDN for icon font

mikewheaton:
This PR removes a temporary Base64-encoded icon font and replaces it with the CDN reference.


philworthington:
Approved

Jahnp:
Approved

PR #665, opened 2016-06-14
Add functionality to people picker

philworthington:
Fixes many of the issues around the current implementation of the PeoplePicker and also includes basic fixes for FacePile. PeoplePicker now shows contextual host and has functionality to add and remove people from the results area. Currently there are conversations on the design side surrounding the PeoplePicker so more design changes and variants could be forthcoming. This PR fixes basic functionality issues.


mikewheaton:
Approved! Builds without errors and works well.

battletoilet:
Approved I had to update some dependencies to support node 6 and fix a weird issue in bundling where there was an extra (). Looks really good man thanks for getting this done.

Jahnp:
@battletoilet --was that causing it to fail? The parens were intentional to make sure the value from that function is assigned immediately.

battletoilet:
Yes, it was causing a failure.

PR #666, opened 2016-06-15
Wdo3650/beta/mdl2/commandbar

wdo3650:
MDL2-ified commandbar with color and behavior changes.

### DIFF
--- a/src/components/CommandBar/CommandBar.scss
+++ b/src/components/CommandBar/CommandBar.scss
@@ -6,8 +6,10 @@
 // Command Bar styles
 
 $CommandBar-height: 40px;
+$CommandBar-md-height: 44px;
 $CommandBar-iconSize: 18px;
 $CommandBar-sideAreaWidthsm: 128px;
+$SearchBox-width: 220px;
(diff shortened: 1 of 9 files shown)


mikewheaton:
Looks like there are some bug fixes in here. Are those specific to the new behaviors for the CommandBar, or something we should be getting fixed in Fabric 3.0 also?

wdo3650:
@mikewheaton yeah, there are fixes for memory leaks. It would be ideal to get the memory leaks fixed in 3.0.

mikewheaton:
@wdo3650 Could you make a quick PR to bring those changes into master, or is there more work involved because of the differences with the MDL2 command bar?

mikewheaton:
The MDL2 icons aren't loading for me. Any ideas?

image

wdo3650:
@mikewheaton the fixes might be a bit more involved. let's assign the bug to me and I can work on it.

wdo3650:
Abandoning this PR. Will open another with SearchBox and CommandBar.

PR #669, opened 2016-06-16
Wdo3650/personacard jeans

wdo3650:
Added fabric.Animate to animate detail box height. Fixed Edge bug and properly removed listeners for bound handlers.


mikewheaton:
Approved! Looks and works great

battletoilet:
Approved

Issue #673, opened 2016-06-20
Errors while building Fabric


pkbullock:
Hi

I have cloned the master repository and followed the instructions on building the project, however I am getting an error when running "gulp" command: "no writecb in Transform class", is this an error with the fabric or local issue do you know?

I have attached a screenshot.

Thanks
Paul
![2016-06-20](https://cloud.githubusercontent.com/assets/8781041/16190451/3e4e4f24-36d7-11e6-89f7-e030b4c84511.png)

mikewheaton:
Thank you for reporting this, @pkbullock. I'm also encountering a bunch of errors and warnings while running `gulp` on a fresh clone of the Fabric repo. Working on it now.

mikewheaton:
I was hoping this was somehow related to the Sass linting errors, but it doesn't appear to be. 

@battletoilet, could you take a look? I'm getting the same errors you see in the screenshot above, and it seems to be related to the timing of the build. If I manually create an empty `dist/css/fabric.components.min.css` file then the build proceeds past the first error, until it starts looking for another nonexistent file:

![image](https://cloud.githubusercontent.com/assets/1382445/16204198/a2ef1dc8-36d2-11e6-8ea8-40ede0511619.png)

mikewheaton:
It's now failing in the Travis build also.

mikewheaton:
Tracked down the issue to an updated version of gulp-header that was released on Friday. PR #676 has been updated to resolve this issue.

mikewheaton:
Fixed by #676

PR #670, opened 2016-06-16
Created stub file for icons adjustments. Added ramp sizes.

wdo3650:
Created stub file for icons adjustments. Added ramp sizes to icons sass file.

PR #676, opened 2016-06-20
Fix Sass linting issues and build error

mikewheaton:
Fix for issue #673 and should allow PR #670 to build without errors.

A recent update to the Sass linter added a new rule that we were failing. All pseudo-elements now have two colons: `::after` and `::before` to differentiate them from pseudo-classes like `:hover`.

There was also a build error (see #673) that turned out to be caused by the latest release of gulp-header. I've updated package.json so that we pull in the latest working version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "gulp-file-include": "^0.13.7",
     "gulp-folders": "^1.1.0",
     "gulp-foreach": "^0.1.0",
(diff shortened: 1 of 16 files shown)


philworthington:
Approved. Daring changes. ⚡ ⚡ ⚡

wdo3650:
Approved.

PR #678, opened 2016-06-21
added icon variables, ramp and adjustments file.

wdo3650:


mikewheaton:
Approved, thanks for catching that it was headed to the wrong branch. Makes sense to include in Fabric 4.0 along with the new icons.

src/sass/_Fabric.Icons.scss
```diff
@@ -347,3 +347,20 @@
     z-index: $ms-zIndex-back;
   }
 }
+
+// Fabric Core Icon variables
+$ms-Icon-12: 12px;
+$ms-Icon-16: 16px;
+$ms-Icon-20: 20px;
```
philworthington: I believe the pattern is to lowercase variable names.
$ms-icon-12
$ms-icon-16
$ms-icon-20

philworthington: Also, we should be using $ms-icon-s, $ms-icon-m, $ms-icon-l since we are prescribing a ramp.

mikewheaton: I'm leaning in this direction as well. The benefit of using 12, 16, and 20 is that if we were to introduce a new size of 18 it would fit right in. The downside is that we're not matching the convention of our other styles, and we're not make even an implicit suggestion of which is the base size. Let's use small, medium, and large and if we do need to add another size in between, we can adjust the ramp to reflect the typical usage.

src/sass/_Fabric.Icons.scss
```diff
@@ -349,18 +349,23 @@
...
-$ms-Icon-16: 16px;
-$ms-Icon-20: 20px;
+$ms-icon-xs: 10px;
+$ms-icon-s: 12px;
+$ms-icon-m: 16px;
+$ms-icon-l: 20px;
 
-.ms-Icon--12 {
-  font-size: $ms-Icon-12;
+.ms-icon-xs {
```
mikewheaton: These should be modifiers, right? E.g. `ms-Icon--xs`

mikewheaton:
Approved 🍦

Jahnp:
Approved

PR #679, opened 2016-06-21
Fix hover issue with placeholder TextField

### LINKED ISSUE #316: TextField: Placeholder interferes with input field hover
The placeholder label prevents the caret from showing up in the input field on hover. @tomaon originally saw this and created a Less fix in issue #267.

mikewheaton:
Fixes #316 

This PR fixes an issue where the label for a placeholder TextField was interfering with the hover. Without IE10 support for `pointer-events: none` I had to instead make the text field transparent and position the label behind it. This is the same method that was originally suggested by @tomaon in #267.

Also removed the background color from the underline variant, so that it can be used on any page without showing a large white box where the field is.


philworthington:
The white box was part of the design when an underline textfield is being used in an environment where edit and read only modes are shown such as in the details pane. The white background makes it clear that the textfield is in edit mode versus a read only mode with a transparent background. I'm not sure if we want our textfields being placed on any background color due to accessibility issues as well.

mikewheaton:
Turns out that the white field I was changing to transparent was for a deprecated style. I've pushed some changes to remove this deprecated style.

philworthington:
Approved 🐤

Jahnp:
Approved

PR #682, opened 2016-06-22
Mistaken code comment in textfield.ts

Rick-Kirkham:
The code DISPLAYS the placeholder text (as it should) when there is no
other value to the field; so the code comment said the opposite of what
the code actually does (and should do).


mikewheaton:
Approved 👍 

Great catch, thanks!

philworthington:
Approved

PR #696, opened 2016-06-29
updated messagebar icons with mdl2.

wdo3650:
FYI, the Blocked2 icon for the blocked variant is missing. An update was made to the cdn for the icons which may take awhile to show up. This should make the icon appear when the cdn updates.


mikewheaton:
This PR uses a new icon that is not yet available on the CDN. It's not ready to merge until the CDN gets updated.

mikewheaton:
I've opened #738 to track the issue with the icon font missing the `Blocked2` icon.

PR #715, opened 2016-07-10
Tidy components' JSON and remove old template files

mikewheaton:
This PR removes unused data from the component JSON files and also removes three old HTML template files.


src/components/Button/Button.json
```diff
@@ -1,16 +1,6 @@
...
-  "fileOrder": [
-    "Button.html",
-    "Button.Primary.html",
-    "Button.Hero.html",
-    "Button.Compound.html",
-    "Button.Submenu.html"
-  ],
-  "props":  {
+  "props": {
     "label": "Create Account",
```
mikewheaton: @battletoilet: All of the props should be coming from the documentation files, right? When I remove these props entirely I get build errors. The same goes for CommandButton, MessageBanner, and SearchBox.

PR #719, opened 2016-07-12
Documentation cleanup

mikewheaton:
This PR removes outdated usage instructions, updates the dependencies for each component based on its JSON file, and adds a plugin to output HTML that can be referenced and copy-pasted.


battletoilet:
It looks like we would still need to add the implementation code snippets for all the components.

mikewheaton:
Yes, this includes the code snippet for Breadcrumb but the others haven't been added yet.

PR #749, opened 2016-07-28
Update theme colors

mikewheaton:
This PR updates three of the theme colors ever-so-slightly to match the latest design spec.


mikewheaton:
I'm not sure why PullApprove isn't working on this PR. It must know that this is a zero-risk PR that is fine to go ahead and merge. 😉

PR #755, opened 2016-07-30
Remove components

### LINKED ISSUE #753: Remove components and related build processes
With [Office UI Fabric JS](https://github.com/OfficeDev/office-ui-fabric-js) public, we can now remove all of the components from this 'core' project. Fabric JS will then use an NPM package to bring in the core styles from this repo.

mikewheaton:
Fixes #753 

This PR removes the components, as they have been moved to [Fabric JS](https://github.com/OfficeDev/office-ui-fabric-js/).

It's a very quick, rough job that builds successfully and gets the 'dist' folder looking right, but there are still remnants in the build system that need cleaned up. We can handle the final cleanup in a separate issue – the goal here was just to quickly remove the duplicate components to make sure people are finding them only in the [Fabric JS](https://github.com/OfficeDev/office-ui-fabric-js/) repo.


battletoilet:
I went through and removed the rest of the component build code and the samples. I can back out the sample deletion but i think it would make sense to just get full documentation going with colors, type ramp, icons etc.

mikewheaton:
Wow, thanks @battletoilet! I agree that we should take out the samples as well, since those are really to demonstrate the components. As part of the site updates we can look into having demonstration pages (for colors, icons, etc.) as part of this repo, which could then be pulled into the site. But for now it's best to keep it simple.

mikewheaton:
Approved 🚴 🚴 🚲 🏃

battletoilet:
Approved

PR #758, opened 2016-08-03
Inherit line-height for icons

mikewheaton:
This PR sets line-height to 'inherit', so that icons act like any other text span rather than setting an explicit line height.


atneik:
Approved

mikewheaton:
This may break some of our components, and other apps built with the icons, where adjustments have been made to padding/position to account for the icon's line height. Talking with @philworthington, I'm now not sure it's worth introducing breaking changes for this.

@Jahnp, thoughts?

philworthington:
We got it sorted. Approved

philworthington:
Approved. We got it sorted.

PR #766, opened 2016-08-09
Removing the colors from all the font mixins.

### LINKED ISSUE #767: Remove all colors from font classes/mixins.
The font mixins have color in them. It causes a lot of issues, please let's consider removing them. Otherwise even just consuming the mixins (not the classes) is difficult and we end up duplicating them elsewhere.

PR here:
https://github.com/OfficeDev/office-ui-fabric-core/pull/766

dzearing:
Fixes #767 

We hit numerous issues when using ms-font-X on anything because it hardcodes the foreground color, rather than allowing the use case to specify. This creates all sorts of nasty problems trying to get colors to work because you end up making rules like:

.odoMyFooLink.ms-Link.ms-font-s:visited {
color: red;
}

Which totally breaks when someone changes the font to be regular...


atneik:
Approved. We should also look for any other major updates to the core, as we could incorporate them with MDL2 breaking updates all over.

mikewheaton:
Before we can merge this we need to update the classes (e.g. `.ms-font-m` to include color), even though the mixins will not include these colors.

Jahnp:
We actually should remove the colors from the classes as well—it’d be confusing for designer/developer communication if the properties of same-named classes and mixins differ that drastically.

mikewheaton:
I thought we agreed that these colors are useful to help developers quickly style body text, and that we want to encourage them to use the correct color.

mikewheaton:
Although now that I'm seeing concerns from SharePoint where these classes aren't compatible with theming, maybe we do want to remove the color after all. We should just consider recommending the standard text color somewhere in our documentation.

ericthompson:
I agree - documentation is key as the `ms-color-neutralPrimary` *is* a part of the design language.

Jahnp:
Yes--where the discussion netted out was that it would make more sense for the classes to match the mixins, but recommend or provide a mechanism (e.g. a `.ms-Fabric` class like in Fabric-React) for setting a base text color at the document level. That was the original spirit of including color in the classes: to ensure that there was an intentional, designed baseline text color that all elements of the type ramp snap to. Setting a document-level text color would do away with the override issues affecting SP theming while still ensuring the right colors are used for plain text.

This would *definitely* need to be reflected on the site--probably both in the typography section as well as the starter template.

mikewheaton:
Approved! Thanks for the discussion here. 😄

mikewheaton:
@dzearing, could you merge in the latest changes from `master` and then merge in this PR? Thanks!

dzearing:
Working on it

PR #771, opened 2016-08-13
Adds PACKAGES.md for package mgmt info

ericthompson:


ghdocs/PACKAGES.md
```diff
@@ -0,0 +1,21 @@
...
+###Install with Bower
+To install Fabric using Bower, run the following command:
+```
+$ bower install office-ui-fabric
+```
+
+###Install with npm
+To install Fabric using npm, run the following command:
+```
+$ npm install office-ui-fabric
```
Jahnp: I don't think this name will be accurate for very long, right?

PR #793, opened 2016-09-02
Re-adding variables fro high contrast links

antonlabunets:
95fcf8813cc64a0a2828b74b40630daa17b39969 made the white on black link fail the contrast requirements. So fixing that and adding back the missing colors.


antonlabunets:
Abandoning since users should be able to override.

Issue #805, opened 2016-09-22
Customization of colors to match existing SharePoint branding


lafe:
We are using Office UI Fabric to build a SharePoint Add-In which are deployed mostly on premises. As discussed in #783 SharePoint themes are currently not supported. Unfortunately, the customer uses a customized branding (mostly slight color corrections to match it with their own corporate identity).
The customer has the expectation that the new Add-In should match the branding of the existing SharePoint pages. We are currently looking into ways to customize the colors in Office UI Fabric in a way that is compatible with future releases of the Office UI Fabric but were unsuccessful so far. For example, Bootstrap uses the [!default](https://robots.thoughtbot.com/sass-default) identifier to define its own variables, so that it is pretty easy to create a customized experience.
Is there any similar mechanism for Office UI Fabric? Would a pull request that enables a similar behavior to Bootstrap be successful or are those customizations not supported?

ericthompson:
At the moment, we don't have a clear picture of how the Fabric projects interact with SharePoint theming. We've started down the path of supporting this for [Fabric React](https://github.com/OfficeDev/office-ui-fabric-react), but we're still in the early stages of bringing this together.

We're looking into this, and hope to provide an update.

PR #815, opened 2016-10-06
Added !default to variables

lafe:
As discussed in issue #805, Office UI Fabric currently does not support customization, which is necessary if customers have their own branding. 
This pull requests adds the [`!default`](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#variable_defaults_) flag to the variables defined in the Office UI Fabric. This flag tells SASS to leave the value of a variable unchanged, if it is already defined. With this approach, you can define your own values in a separate file and include them in the build process of your custom Office UI Fabric. Only those values will be set to their default values, that have not been defined previously in your own variables file.

Your custom SASS file might look like the following:
``` SASS
@import "fabricVariables"; //Includes your variables
@import "fabric"; //Main Fabric.scss file (located elsewhere, included in the sass build process)
@import "fabric.components"; //Main SCSS for components (located elsewhere, included in the sass build process)

//Your additional customizations
```

### DIFF
--- a/src/sass/_Fabric.Animations.scss
+++ b/src/sass/_Fabric.Animations.scss
@@ -10,12 +10,12 @@
 // The original class names are deprecated and will be removed in a future release.
 
 // Variables
-$ms-ease1:     cubic-bezier(0.1,0.9,0.2,1);
-$ms-ease2:     cubic-bezier(0.1,0.25,0.75,0.9);
-$ms-duration1: 0.167s;
-$ms-duration2: 0.267s;
-$ms-duration3: 0.367s;
-$ms-duration4: 0.467s;
+$ms-ease1:     cubic-bezier(0.1,0.9,0.2,1) !default;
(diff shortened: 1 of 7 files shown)


mikewheaton:
Approved. This is a great idea, as developers can override variables without actually modifying Fabric itself.

Jahnp:
Approved. Super super cool--thanks for the contribution @lafe! It does look like there are conflicts with `master` now, though--once those have been resolved we'd be happy to merge this in.

lafe:
After merging the current master branch in my local repository, I was unable to spot any conflicts. Please let me know, if there are specific problems you encounter.

Jahnp:
Strange, that didn't seem to resolve the conflicts that GH was reporting. I've updated the PR and verified that your changes are all still there, so once CI finishes I'll go ahead and merge this in. Thanks again for the contribution @lafe!

PR #823, opened 2016-10-17
Add BrandIcon component

leddie24:
Adds smart control media queries for PNG brand icons.  Added brand icon classes to Fabric.Brand.Icons.Output.scss.

### DIFF
--- a/src/sass/Fabric.Brand.Icons.Output.scss
+++ b/src/sass/Fabric.Brand.Icons.Output.scss
@@ -1,0 +1,82 @@
+// Images Path
+$imagesPath: "https://appsforoffice.microsoft.com/fabric/assets/Product/PNG";
+
+// Icon Names
+$iconList: Access Excel InfoPath Office OneDrive OneNote Outlook PowerPoint Project SharePoint Visio Word;
+
+// Icon Sizes
+$iconSizes: 16 48 96;
+
+// Mixin to loop through iconList and generate classes
+@mixin iconClass($size, $retina) {
+  @each $name in $iconList {
+    &.ms-Brand--#{$name}#{$size} {
+        @include Icon($name, $retina, $size);
+    }
+  }
+}
+
+// Generate icon size classes
+@mixin IconSizeClass() {
+	@each $size in $iconSizes {
+		&.ms-Brand--Icon#{$size} {
+			@include IconSize($size);
+		}
+	}
+}
+
+// Set image container width and height
+@mixin IconSize($size) {
+	width: #{$size}px;
+	height: #{$size}px;
+}
+
+// Set image source and size.
+@mixin Icon($name, $retina, $size) {
+	background-image: url(#{$imagesPath}/#{$name}_#{$size}x#{$retina}.png);
+}
+
+// Default icon set for Pixel Ratio 1
+.ms-Brand {
+	background-size: 100% 100%;
+	@include IconSizeClass();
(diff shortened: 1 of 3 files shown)


ericthompson:
I may be jumping on this a bit early, but where did we want the documentation for this to go? Just under /ghdocs?

leddie24:
I added documentation in fabric-website.  There's a PR for it here

https://onedrive.visualstudio.com/Design/Design%20Development/_git/fabric-website/pullrequest/30708?_a=overview

Should we add another page inside of /ghdocs for brandicons, or add content to /src/documentation/pages/icons?

ericthompson:
@leddie24 - sorry, forgot about the site! No worries, I'll take a look at the site - just missed the notification 😪

src/sass/Fabric.Brand.Icons.Output.scss
```diff
@@ -0,0 +1,82 @@
...
+@mixin iconClass($size, $retina) {
+  @each $name in $iconList {
+    &.ms-Brand--#{$name}#{$size} {
+        @include Icon($name, $retina, $size);
+    }
+  }
+}
+
+// Generate icon size classes
+@mixin IconSizeClass() {
```
mikewheaton: Let's put the mixins and variables in `_Fabric.Brand.Icons.scss` so that these can be used in other apps without outputting any CSS.

src/sass/Fabric.Brand.Icons.Output.scss
```diff
@@ -0,0 +1,82 @@
...
+	@each $size in $iconSizes {
```
mikewheaton: Change tabs to spaces. Maybe we can make an editor config file so this is the default?

src/sass/Fabric.Brand.Icons.Output.scss
```diff
@@ -0,0 +1,82 @@
+// Images Path
+$imagesPath: "https://appsforoffice.microsoft.com/fabric/assets/Product/PNG";
```
mikewheaton: Did we decide against doing this for the document type icons as well?

src/sass/Fabric.Brand.Icons.Output.scss
```diff
@@ -0,0 +1,82 @@
...
+.ms-Brand {
+	background-size: 100% 100%;
+	@include IconSizeClass();
+	@include iconClass(16, 1);
+	@include iconClass(48, 1);
+	@include iconClass(96, 1);
+}
+
+// Pixel Ratio 1.5
+@media only screen and (min--moz-device-pixel-ratio: 1.5),
```
mikewheaton: I did a little research and it looks like [`min-resolution` is the standard](https://developer.mozilla.org/en-US/docs/Web/CSS/resolution). For the [widest support](http://caniuse.com/#feat=css-media-resolution) you can use DPI units, where 96dpi is the equivalent of 1x, 144dpi is 1.5x, and so on.

So these multiple media queries with all of the browser prefixes can be replaced with a single:

`@media only screen and (min-resolution: 144dpi)`

Which our autoprefixer process, to make it work on Safari, outputs as:

`@media only screen and (-webkit-min-device-pixel-ratio: 1.5), only screen and (min-resolution: 144dpi)`

mikewheaton:
Approved!

ericthompson:
Approved!

PR #825, opened 2016-10-21
Make the grid cell mix-in be with @media info.

lijunle:
- These mix-in can be re-used as _right_ mixin now.

### DIFF
--- a/src/sass/Fabric.Responsive.Utilities.Output.scss
+++ b/src/sass/Fabric.Responsive.Utilities.Output.scss
@@ -205,40 +205,40 @@
 
 // Small Offset classes 0 ... 11
 .ms-u-smOffset11 {
-  margin-left: 91.66666666666666%;
+  @include ms-u-smOffset11;
 }
 .ms-u-smOffset10 {
-  margin-left: 83.33333333333334%;
+  @include ms-u-smOffset10;
 }
 .ms-u-smOffset9 {
-  margin-left: 75%;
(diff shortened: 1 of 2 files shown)


src/sass/Fabric.Responsive.Utilities.Output.scss
```diff
@@ -242,781 +242,771 @@
...
+.ms-u-mdPush1 {
+  @include ms-u-mdPush1;
+}
+.ms-u-mdPush0 {
+  @include ms-u-mdPush0;
+}
+
+// Medium Offset classes 0 ... 11
+.ms-u-mdOffset11 {
+  margin-left: 91.66666666666666%;
```
Jahnp: Is there a reason there aren't mixins for these classes as well? If we're removing the higher-order media queries, these should be accounted for as well.

lijunle: Sorry, this is my mistake. I am going to fix this.

lijunle:
Hi, @Jahnp Thanks for reviewing! I have updated the PR to create the correpsonding mix-ins for the offset styles.

Jahnp:
Approved

Awesome! Thanks so much @lijunle! Looks great now.

mikewheaton:
Approved. Thanks @lijunle!

lijunle:
Thank you all for reviewing! 👍

PR #842, opened 2016-11-17
Update CDN paths to support HTTP/2

mikewheaton:
The new CDN path supports the [HTTP/2](https://en.wikipedia.org/wiki/HTTP/2) protocol, which allows for loading multiple files in a single request. This improves performance when requesting multiple files from the same server.

The old CDN paths will continue to work (they point to the same server) so this is not a breaking change.


Jahnp:
Approved

This is *super* cool. I didn't know the CDN allowed that. Thanks for this @mikewheaton!

mikewheaton:
Approved

PR #858, opened 2016-11-30
Updated with 2016.11.23 icons

micahgodbolt:


src/sass/Fabric.Icons.Output.scss
```diff
@@ -34,6 +34,10 @@
   @include ms-Icon--l;
 }
 
+// @todo: 'DynamicsCRMLogo' class is deprecated, will be removed in a future release
+// @todo: 'CarotRightSolid8' class is deprecated, will be removed in a future release
```
mikewheaton: I see you've moved the comments up here, but where are the deprecated classes themselves?

micahgodbolt: DynamicsCRMLogo is still down there. CarotRightSolid8 is a bit confusing though, as the original code simply had CarotRightSolid8 twice. It wasn't deprecated for something else.

src/sass/Fabric.Icons.Output.scss
```diff
@@ -578,9 +586,7 @@
 .ms-Icon--CaretUp8::before { @include ms-Icon--CaretUp8; }
 .ms-Icon--CaretDown8::before { @include ms-Icon--CaretDown8; }
 .ms-Icon--CaretLeftSolid8::before { @include ms-Icon--CaretLeftSolid8; }
-// @todo: 'CarotRightSolid8' class is deprecated, will be removed in a future release
-.ms-Icon--CarotRightSolid8::before,
```
mikewheaton: This class shouldn't be removed, just deprecated.

src/sass/Fabric.Icons.Output.scss
mikewheaton: It was a typo. The old class was 'carot' with an O, and the new is 'caret' with an E.

mikewheaton: Did you want to move both the comment and the deprecated selector to the top? I think you wanted to be able to copy the newly generated classes over top without losing anything, right?

micahgodbolt: @mikewheaton alright. Moved those 2 deprecated classes up to the top

mikewheaton:
Approved 🌮

mikewheaton:
The updated icon font has been deployed to the CDN, so we're good to merge this as soon as it gets one more review. 😄

ericthompson:
Approved! 💃

PR #860, opened 2016-12-02
Add support for scoped versioning

Jahnp:
# Version scoping
This pull request adds support for "version-scoping" Fabric Core, which effectively wraps or "scopes" every Fabric class under a modifier of the `.ms-Fabric` base class, which is tied to the current version. If the version is `5.0.1`, this "scope class" would take the form `.ms-Fabric--v5-0-1`.

## Why is this useful?
Version scoping is targeted at the scenario where multiple versions of Fabric Core may live on a single web page, whose differences between major versions could potentially introduce regressions or conflicts.

A practical example of this is the SharePoint Web Parts used for publishing articles. Web Parts are effectively miniature applications, each of which can depend on their own version of Fabric. To ensure that they can continue to exist on the same publishing page without:
1. Risking regressions from other web parts that might be using an older version of Fabric, and 
2. Forcing them to consume the latest (which would likely require work to update), 

the Web Part developer can scope the Fabric styles for their entire Web Part to the particular version that was used during development.  

## Usage
Usage of version scoping is straightforward. Simply append the "versioned class" to the container element you wish to scope, then use Fabric Core classes as you normally would.

Here's a minimal HTML page demonstrating this: 

```html
<html>
    <head>
        <!-- Link to the scoped styles for the current version of Fabric Core -->
        <link rel="stylesheet" href="/fabric-5.0.1.scoped.css">
    </head>
    <body>
        <!-- Add the scoping class -->
        <div class="ms-Fabric--v5-0-1">
            <!-- 
                Use Fabric as usual. These styles will be unaffected
                by other versions of Fabric. They also cannot be used
                outside of a scoped class.
            -->
            <h1 class="ms-font-su">Some scoped title <i class="ms-Icon ms-Icon--Info"></i></h1>
        </div>
    </body>
</html>
```

## Notes on fonts
Icon `@font-face` definitions (including the `font-family` name) and their font files are only scoped to **major versions** of Fabric. What this means is that all of the normal, additive changes to icon font files that occur within a major version's lifetime will be applied to a single font file, rather than maintaining separate font files for each minor or patch release. What this means is that there will not be `fabricmdl2icons-5.1.0.ttf`, `fabricmdl2icons-5.2.0.ttf`, etc. There will simply be `fabricmdl2icons-5.ttf`.

This is because generally icon font changes are purely additive, and rarely change significantly even between minor versions. However, the CSS classes for those icons (e.g. `ms-Icon--X`) *are* version scoped like normal classes. Also, note that the "unscoped" version of the icon font file as used today will still be available.

Finally, the standard, "unscoped" `@font-face` definitions to the Segoe UI webfonts are included in scoped Fabric unchanged. These fonts almost never change appreciably and are not considered risky to depend on between major releases.


src/documentation/pages/Scoping/index.html
```diff
@@ -0,0 +1,124 @@
+@@include('../../templates/modules/components/DocumentationPageHeader.html')
```
Jahnp: These demo files will probably need to be cleaned up/refactored a bit, since I'd rather not provide a separate file to "mock" potential conflicts.

src/samples/Scoping/index.html
```diff
@@ -0,0 +1,61 @@
...
+  <main class="ms-Grid">
+    <div class="ms-Grid-row">
+      <div class="ms-Grid-col ms-u-sm10 ms-u-smPush1 ms-u-smPull1">
+        <h2 class="ms-font-xxl">Version 4</h2>
+      </div>
+    </div>
+
+    <div class="ms-Grid-row">
+      <div class="ms-Grid-col ms-u-sm10 ms-u-smPush1 ms-u-smPull1">
+        <h2 class="ms-font-xxl">Version 4</h2>
```
mikewheaton: I'm not seeing the scope classes in use on this page. Did you want to show an example of how v4 and v5 could coexist with different styles?

Jahnp:
I'm going to go ahead and close this for now to deal with some of the cleanup required from merging with `master`. Will reopen soon.

PR #868, opened 2016-12-11
Restructure, tidy, and bug fix SCSS

### LINKED ISSUE #349: Output.scss files in Fabric?
What's the story with all of these Output.scss files in Fabric? They were added back in December with a commit message of "Adds sass converted files to fabric". Is this some relic of when the project was in LESS but some teams were needing Sass variables?

The result is that we have a file like `_Fabric.Icons.scss` with a mixin for every icon, and these mixins are used nowhere else but `Fabric.Icons.Output.scss`. Seems unnecessary and hard to maintain. But I might be missing something here...

image

### LINKED ISSUE #832: Scope all mixins and variables, and name them consistently
Here are some mixins we currently have:

```
ms-u-borderBox
ms-baseFont
ms-Grid
high-contrast
ms-Icon--s
brandIconClasses
```

We should prefix all of our mixins with `ms-` so that they can be safely used in other projects without conflicts, and settle on whether we want to use camelCase or kebab-case. Thoughts?

### LINKED ISSUE #865: Language font stack overrides missing from fabric.css
The language font stack overrides (`Fabric.Typography.Language.Overrides.Output.scss`) are only imported by [Fabric.RTL.scss](https://github.com/OfficeDev/office-ui-fabric-core/blob/master/src/sass/Fabric.RTL.scss) and not [Fabric.css](https://github.com/OfficeDev/office-ui-fabric-core/blob/master/src/sass/Fabric.scss). This means that languages that are read left-to-right that we have special font stacks for (e.g. Hindi, Chinese, Turkish) have not been displayed in the correct font.

I'm working on a large PR that will include a fix for this.

### LINKED ISSUE #866: Brand Icons missing from fabric.rtl.css
We forgot to add a reference to `Fabric.Brand.Icons.Output` in [`Fabric.RTL.scss`](https://github.com/OfficeDev/office-ui-fabric-core/blob/master/src/sass/Fabric.RTL.scss). I'll be fixing this, and making sure LTR and RTL don't contain so many duplicated references, in a PR I'm working on now.

mikewheaton:
This PR restructures the `sass` folder to remove the confusing "output" file names (fixes #349), giving us a new structure like so:

```
- sass
  - mixins
    - _Animation.Mixins.scss
    ...
  - variables 
    - _Animation.Variables.scss
    ...
  _Animation.scss
  Fabric.scss
  ...
```

All of the variables and mixins are clearly separated out from the partials that actually generate CSS. Those partials use the underscore naming convention (e.g. `_Grid.scss`) to indicate that they shouldn't be converted directly to CSS. The only files without the underscore are `Fabric.scss` and `Fabric.RTL.scss`.

Along the way, I discovered several other bugs and inconsistencies, many of which are fixed by this PR. Here are the changes and fixes:

## No impact on output CSS or use of mixins/partials
- Standardized to use single quotes everywhere.
- Removed `_Fabric.i18n.scss`. We replaced this with another file a while back, now located in `mixins/_Directionality.Mixins.scss`, but this original hadn't been deleted. Fortunately, it was loaded before the proper mixins and had no effect on the output.
- Added missing semicolons, fixed indentation, etc.
- Renamed the animation `slideLeft400` to `slideLeftIn400` to match the others. This is a change only in the `@keyframes` name and does not impact the mixins or class names.
- `Fabric.RTL.scss` now simply imports `Fabric.scss` and then references RTL-specific overrides. Reduces the chance of future bugs caused by adding a reference to one file and not the other.

## Changes to mixins/partials
- Many mixins/partials have been renamed, moved, merged together, etc. This will be a breaking change for anyone importing our SCSS files into their project. However, the improved structure and naming should make it easy to update references.
- Removed some mixins that created browser prefixes for animations. This is handled by Autoprefixer so the mixins are not necessary.
- Added a new mixin, `ms-animation-keyframes()`, that outputs all of the `@keyframes` used by the animation classes. Previously, you could use mixins to add animation to an element, but that animation wouldn't work because the `@keyframes` were unavailable.
- Renamed the `animationMix()` mixin to `ms-animation()`. I'll be renaming more mixins and variables in a future PR to fix #832, as we're inconsistent here and should be scoping them all.
- Renamed `_Fabric.Base.scss` to `_Wrapper.scss`, so we can consistently refer to it as the 'wrapper' class. No change to the actual output of `ms-Fabric`.
- Renamed `_Fabric.Common.scss` to `_References.scss` and made sure this contains all of the variables and mixins. Then imported this file into `Fabric.scss` so that no references are missed.
- The `ms-Grid-col()` mixin now styles child `.ms-Grid` classes, so that grids can be nested correctly.
- Renamed the `Typography` files to `Font` to match the actual variable/mixin/class names.

## Differences in the output CSS

- Animation classes no longer contain browser prefixes. These are not needed for the browsers we support. If prefixes are needed in the future, this will be taken care of by Autoprefixer.
- `fabric.css` now contains the language overrides for font stacks (fixes #865).
- `fabric.rtl.css` contained many animation classes that were referring to `@keyframes` by the wrong name. This meant that the animations weren't working in RTL. They now refer to the correct animations: `slideLeftIn10`, `slideLeftIn20`, `slideLeftIn40`, `slideRightIn10`, `slideRightIn20`, `slideRightIn40`, and `slideRight400`.
- `fabric.rtl.css` now includes the BrandIcon classes (fixes #866).
- Fixed mistakes in the animation classes:
    - Fade animations now use the duration specified by the class name (e.g. `ms-u-fadeIn100` lasts 100ms instead of 367ms).
    - ExpandCollapse animations now use the duration specified by the class name (e.g. `ms-u-expandCollapse400` lasts 400ms instead of 367ms).
    - Delay classes now set a delay based on the class name (e.g. `ms-u-delay200` delays for 200ms instead of 267ms).

With so many file changes it's hard to look at the diffs, so I've used the super-helpful [Alan's Compare CSS tool](http://www.alanhart.co.uk/tools/compare-css.php) to compare the master versions of `fabric.css` and `fabric.rtl.css` with the versions from this branch. I went through several times to be sure I'm not introducing any bugs.


micahgodbolt:
I still believe file names should include their category (animation-variable,animation-mixins, animation-classes etc). It'll help a ton in fuzy searching and code review.

mikewheaton:
@micahgodbolt: So your preference on file names would be `_Animation.Variables.scss`, `_Animation.Mixins.scss`, and `_Animation.scss`? I can get behind that. 😄  Beats having three identical file names.

Jahnp:
Agreed, it's useful to distinguish filenames when doing fuzzy search in editors: 
image

Or in open tabs:
image

Jahnp:
Thanks for doing this @mikewheaton! I'll pull down your branch to check out your changes soon. **So** excited to finally get these cleaned up! Things were starting to get huge and crusty.

src/sass/mixins/_Animation.scss
```diff
@@ -0,0 +1,345 @@
...
+}
+
+@mixin ms-u-rotateN90deg {
+  @include ms-animation(rotateN90, $ms-duration1, $ms-ease2);
+}
+
+
+// Fade
+@mixin ms-u-fadeIn100 {
+  animation-duration: 0.1s;
```
Jahnp: Totally understand the motivation here and agree that the original name/value mismatch for these was confusing, but I'm quite sure at the time the values were intentional/by design. While it's probably fine to change them to make more sense, please check with motion design to verify. 

Frankly, I'd prefer we use simple, predictable durations in sensible increments like you've proposed here.

mikewheaton:
@micahgodbolt @Jahnp: I'd also like feedback on `_References.scss`, which used to be `_Fabric.Common.scss`. Is the name an improvement? My concern was that "common" suggested these were popular parts of Fabric, when it's just imports for all of the variables and mixins.

Would it make sense to see something like `@import 'office-ui-fabric-core/References';` at the top of an SCSS file?

mikewheaton:
@Jahnp: The underscore in Sass/SCSS [indicates a partial](http://sass-lang.com/guide#topic-4), which is any file that is included by others and doesn't directly output a CSS file. Our build system doesn't care about the underscore, so it really makes no difference, but I wanted to follow this convention that others might recognize. It's also followed by [Bootstrap](https://github.com/twbs/bootstrap/tree/v4-dev/scss) and [Foundation](https://github.com/zurb/foundation-sites/tree/develop/scss).

Jahnp:
@mikewheaton: I see--so it prevents there from being a `Grid.css` file being generated, but doesn't preclude a `_Grid.css` file from still producing output. Makes sense, I had forgotten that distinction.

mikewheaton:
I've updated the variable and mixin file names to include their type (e.g. `_Responsive.Variables.scss`).

Jahnp:
@mikewheaton I also hear you on the reasoning behind the naming of `_References.scss`. My only "concern" would be that "Common" seems to be a, well, common name for files that import mixins and variables used across a project. I see this internally and externally on a number of projects, but if I'm honest, I don't think it's a big deal.

I think the biggest thing would be to just make sure we call out the breaking change, because I know there are many projects that would be affected by this.

mikewheaton:
@Jahnp: I'm not strongly against "Common", I just preferred "References" a bit more. It will be a breaking change in any case because the old files were all prefixed with "Fabric", so it's a good opportunity to change the name if we want to.

Jahnp:
Approved

Thanks again for doing this @mikewheaton!

micahgodbolt:
@mikewheaton sweet work here! Looking forward to seeing this repo being soooo much better organized.

src/sass/variables/_Font.Variables.scss
```diff
@@ -0,0 +1,52 @@
...
+$ms-font-path-cyrillic:     'segoeui-cyrillic' !default;
+$ms-font-path-easteuropean: 'segoeui-easteuropean' !default;
+$ms-font-path-greek:        'segoeui-greek' !default;
+$ms-font-path-hebrew:       'segoeui-hebrew' !default;
+$ms-font-path-vietnamese:   'segoeui-vietnamese' !default;
+$ms-font-path-westeuropean: 'segoeui-westeuropean' !default;
+
+// Base font stacks
+$ms-font-system-base: 'Segoe UI', -apple-system, BlinkMacSystemFont, 'Roboto', 'Helvetica Neue', sans-serif !default;
+$ms-font-family-base: 'Segoe UI WestEuropean', $ms-font-system-base !default;
```
mikewheaton: This will be investigated more closely in #382. Do you remember if 'Segoe UI WestEuropean' is supposed to refer to a locally-installed font? We have a web font for it on the CDN, but I can't find a corresponding `@font-face` declaration that makes this available.

src/sass/mixins/_BrandIcon.Mixins.scss
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE in the project root for license information.
+
+// Generate classes for product and document icons
+@mixin brandIconClasses($retina) {
+	@each $icon in $ms-productIcons {
```
mikewheaton: Good catch. Looks like this file is using tabs instead of spaces. We're going to be adding linting (#869) soon to keep inconsistencies like this from getting checked in.

src/sass/variables/_Font.Variables.scss
notandrew: No, 'Segoe UI' is the local system font ($ms-font-system-base fonts are all local system fonts). 'Segoe UI West European' is the friendly name for the web font. I'm assuming 'Segoe UI WestEuropean' would fail for Macs and non-Windows devices.

mikewheaton: The web font is spelled as 'WestEuropean': https://github.com/OfficeDev/office-ui-fabric-core/blob/master/src/sass/mixins/_Font.Mixins.scss#L141

I'm going to open a new issue to include "Web Font" in the name of any `@font-face` we declare. It's difficult to tell right now where we're expecting a system font to load and where it's a web font we've defined, which we can name whatever we like. 😄

src/documentation/sass/docs.scss
```diff
@@ -85,7 +85,7 @@ $high-contrast-green: #3ff23f;
 
 // slideLeftIn80
 @mixin ms-u-slideLeftIn80 {
-    @include animationMix((fadeIn, slideLeftIn80), $ms-duration3, $ms-ease1);
+    @include ms-animation((fadeIn, slideLeftIn80), $ms-duration3, $ms-ease1);
```
elisechant: to me this namespacing looks like a vendor prefix and makes it more vague

PR #880, opened 2016-12-20
Revert animation durations

mikewheaton:
Reverts the durations for fade, rotate, and delay animations to what was in Fabric Core 5.1.0. In cases where we previously specified different values for `-webkit-animation-duration` and `animation-duration`, I've used the non-prefixed value.

There is design work happening now to redefine the animations, and we don't want to introduce multiple rounds of breaking changes here. I've opened #881 to track that work.


Jahnp:
Approved

Thanks for this, @mikewheaton.

PR #885, opened 2017-01-03
Automate Icon File Generation (not ready for merge)

micahgodbolt:
Seeing as we have 3 separate files that are just different forms of the same data, it seemed to make sense to generate all 3 files from a single set of data. 

This PR still needs to update Gulp to properly create the temp files before Sass is compiled.


src/icons/data/icons.json
```diff
@@ -0,0 +1,708 @@
+{
```
mikewheaton: This file should now be in the 'dist' folder, right?

micahgodbolt:
{"name":"{name}", "value":"{value}"},

gulp/modules/Plugins.js
```diff
@@ -35,6 +35,8 @@ var Plugins = function() {
 	this.path = require('path');
 	this.pkg = require('../../package.json');
 	this.fs = require('fs');
+	this.yaml = require('js-yaml');
+
```
battletoilet: nit: extra line and if we are not using this anymore let's remove it.

package.json
```diff
@@ -41,6 +41,7 @@
     "gulp-tslint": "^7.0.1",
     "gulp-typescript": "^3.1.1",
     "gulp-util": "^3.0.1",
+    "js-yaml": "^3.7.0",
```
battletoilet: Same here let's remove this if it's not being used.

gulp/modules/Plugins.js
micahgodbolt: good catch

mikewheaton:
Closing this for now as it's not being actively worked on.

PR #887, opened 2017-01-03
Consistent scope and naming for variables, mixins, and keyframes

### LINKED ISSUE #832: Scope all mixins and variables, and name them consistently

mikewheaton:
Fixes #832.

This PR adds scoping (via the `ms-` prefix) and makes the naming of mixins, variables, and keyframes more consistent.

With the exception of the brand icon modifier classes (see below), there are **no other changes to the CSS classes in Fabric**, so while this is a breaking change it won't affect the majority of applications using Fabric as a simple CSS include.

## New naming convention
- Everything in the global scope is prefixed with "ms-".
- CSS classes remain unchanged (following [SUIT CSS conventions](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md)).
    - `ms-Grid-col`
    - _Exception: Pascal case is used for the icon names:_ `ms-Icon--EraseTool`
- Animation keyframes all match their corresponding CSS class.
    - `@keyframes ms-slideRightIn10`
- Variables are kebab case.
    - `$ms-animation-duration-3`
    - `$ms-screen-min-xl`
    - _Exception: camel case is used for colors to match class names:_ `$ms-color-themeLighterAlt`
- Mixins are usually kebab case.
    - `ms-focus-border()`
    - `ms-drop-shadow()`
    - _Exception: If the mixin maps directly to a CSS class, we match the class name:_
        - `ms-u-borderBox()` outputs to `.ms-u-borderBox`
        - `ms-fontSize-s()` outputs to `.ms-fontSize-s`

### Animation keyframes
- Animation keyframes are now prefixed: `slideRightIn10` to `ms-slideRightIn10`

### Classes
- Brand icon modifiers are now lowercase: `ms-BrandIcon--Icon16` to `ms-BrandIcon--icon16`

### Mixins
- **Prefixed all mixins**: `reset-padding` to `ms-reset-padding`
- Brand icons
    - Renamed mixins to kebab case: `brandIconClasses` to `ms-brand-icon-classes`
- Font
    - Scoped all mixins: `SegoeUIWestEuropeanLight` to `ms-font-family-segoeUIWestEuropeanLight`
    - Renamed `language-override-system-fonts` to `ms-font-family-override`
    - Renamed `ms-baseFont` to `ms-base-font`
- Directionality
    - Renamed `baseRtl` to `ms-base-RTL`
- General
    - Renamed mixins to kebab case: `resetAnimation` to `ms-reset-animation`
- Utility
    - Renamed mixins to kebab case: `resetMargins` to `ms-reset-margins`

### Variables
- Responsive
    - Changed order of screen size and min/max: `ms-screen-sm-min` to `ms-screen-min-sm`
- Animation
    - Added "animation" to scope: `ms-ease2` to `ms-animation-ease-2`
- Brand icon
    - Switched to kebab case: `ms-productImagesPath` to `ms-product-images-path`
- Font
    - Renamed `ms-font-family-base` to `ms-font-base-family`
    - Renamed `ms-font-system-base` to `ms-font-base-system`


src/sass/mixins/_Directionality.Mixins.scss
```diff
@@ -32,292 +32,292 @@
...
+    @include ms-RTL {
       clear: left;
     }
   } @else {
       clear: $side;
   }
 }
 
-@mixin float($direction) {
+@mixin ms-float($direction) {
```
Jahnp: Hmm... this is going to cause a *lot* of upgrade pain. Are we confident this is the right change to make?

src/sass/mixins/_Directionality.Mixins.scss
```diff
@@ -32,292 +32,292 @@
...
 
-  @include RTL {
+  @include ms-RTL {
     margin: $top $left $bottom $right;
   }
 }
 
-@mixin margin-left($distance) {
-  @include LTR {
+@mixin ms-marginLeft($distance) {
```
Jahnp: I realize the intent here is to be as consistent as possible across the library and to prevent any potential conflicts when using other SASS libraries, but the names of the directionality mixins were meant to match the CSS properties they affect to make their usage as transparent as possible. I won't block on this right now, but I again do worry that this will be a painful upgrade for a lot of users.

mikewheaton: It's certainly a 'nice to have' that all of the mixins follow the same naming convention and not an absolute must. My thinking was that since we're breaking everything by adding the "ms-" prefix anyway, it's a good opportunity to standardize the name.

mikewheaton: Another option could be to update our other mixins (the ones that don't directly output CSS classes, where we match the names) to use kebab-case.

src/sass/mixins/_Directionality.Mixins.scss
mikewheaton: It's never fun when we introduce breaking changes, but it is a simple find-replace in this case. There's no change in functionality so once all of the build errors are caught it's good to go. I'm planning on doing the update in Fabric React and can help if other teams have concerns.

src/sass/mixins/_Directionality.Mixins.scss
mikewheaton: After discussing in person, we're going to go ahead and kebab case all of the mixins while keeping the "ms-" prefix: `ms-margin-left()`

Jahnp:
Approved

PR #888, opened 2017-01-03
Remove component-specific z-indexes

### LINKED ISSUE #873: Remove component z-indexes
[`_ZIndex.Variables.scss`](https://github.com/OfficeDev/office-ui-fabric-core/blob/master/src/sass/variables/_ZIndex.Variables.scss) contains a bunch of 'standard component indexes' that appear to be leftovers from before [Fabric JS](https://github.com/OfficeDev/office-ui-fabric-js/) was split out as a separate project. They shouldn't be in Core.

Here they are:
```
// Standard component indexes
$ms-zIndex-Callout:        $ms-zIndex-1 !default;
$ms-zIndex-ContextualMenu: $ms-zIndex-1 !default;
$ms-zIndex-Overlay:        $ms-zIndex-2 !default;
$ms-zIndex-Panel:          $ms-zIndex-3 !default;
$ms-zIndex-DatePicker:     $ms-zIndex-3 !default;
$ms-zIndex-Dialog:         $ms-zIndex-3 !default;
$ms-zIndex-PeoplePicker:   $ms-zIndex-3 !default;
$ms-zIndex-Dropdown:       $ms-zIndex-4 !default;
```

mikewheaton:
Fixes #873 by removing the component-specific indexes. These should be defined in projects like Fabric React and Fabric JS if they are necessary.


Jahnp:
Approved

micahgodbolt:
Do we know that these variables are not being used in JS/React? Do we need to deprecate them first?

mikewheaton:
We haven't deprecated undocumented functionality like this in the past. I'm not too concerned about it, as these are very easy to copy into other projects (like Fabric JS) if they're actually needed. I'll go check if they're used.

PR #890, opened 2017-01-04
Inline version number instead of file based

micahgodbolt:


src/sass/Fabric.Scoped.scss
```diff
@@ -4,8 +4,12 @@
 // Office UI Fabric
 // Core styles scoped to the current major version of Fabric
 
+// Fabric version gets set on compile
+$ms-fabric-version: $FabricVersion;
+//
```
Jahnp: We aren't doing this elsewhere--please remove.

src/sass/Fabric.Scoped.scss
```diff
@@ -4,8 +4,12 @@
 // Office UI Fabric
 // Core styles scoped to the current major version of Fabric
 
+// Fabric version gets set on compile
```
Jahnp: Could you expand this to say something along the lines of "$FabricVersion becomes a SASS list like [1,2,3] on build"

mikewheaton:
Approved 👍

PR #899, opened 2017-01-07
Improve font localization and add support for Selawik

### LINKED ISSUE #382: Localization: Set font-family with ms-Fabric wrapper

### LINKED ISSUE #852: Add support for Selawik as an alternative to Segoe UI web font
It would be helpful for community if Fabric team can use [Selawik](https://github.com/Microsoft/Selawik) as fallback approach where Segoe license has restriction. There is also a [Symbol](https://github.com/Microsoft/fonts) font which can be used along with Selawik. 

We can have wider adoption of this library and MDL2 outside of Office/MS linked usages. 

Looks like the development on the two project has stalled or not published to GitHub for over a year now. if Fabric team adopts these fonts, their further development can happen. It would be great contribution to open source community by Microsoft

### LINKED ISSUE #871: Document how to override the base font-family to use another font
Separating this work item out from #852. There are limitations on the use of Segoe UI, which prevent developers from using Fabric for projects that aren't specifically for Office, or from hosting their own fonts without relying on the CDN.

The request here is to document how to override the base `font-family` to set a different font stack, which would prevent Fabric from pulling Segoe UI from the CDN as usual.

### LINKED ISSUE #878: Include "Web" in the name of any web fonts
Right now we generate web font definitions with names like "Segoe UI WestEuropean". These are difficult to distinguish from system-installed fonts like "Segoe UI". Whenever we define a new web font with `@font-face`, let's make it clear with a name like "Segoe UI West European Web" so that we can tell what's expected to be local and what's a web font.

### LINKED ISSUE #893: Remove web fonts in TTF format
Our web font definitions (both Segoe UI for text and MDL2 for icons) include TTF references, which aren't necessary for the browsers we support. WOFF actually has better browser support now than TTF, while coming in at roughly half the download size.

mikewheaton:
# Changes
- Font families are now set exclusively on the `ms-Fabric` wrapper class. These overrides are generated using a new `ms-font-family-language-override()` mixin. **This means that the Fabric wrapper is required on all pages.** The benefit is that now that the font family is set in the root, it can cascade correctly into the Fabric React and Fabric JS components. Localized components at last! Fixes #382.
- All of the font definition mixins, such as `ms-font-family-segoeUICyrillicLight()`, have been replaced with a single `ms-font-face()` mixin that generates `@font-face` definitions for the parameters provided.
    - The exception is for Leelawadee UI (Thai), which requires overrides due to it not having "light" or "semibold" weights.
- Removed the `ms-font-family-override()` mixin.
- Added support for the [Selawik web font](https://github.com/Microsoft/Selawik). Fixes #852.
- Removed the references to TTF web font files. Fixes #893.
- All web font names now follow a format like "Segoe UI Web (Greek)" so that it's clear when a web font is being used. Fixes #878.
- Local fonts are no longer referenced with the `@font-face` definitions. For clarity, local fonts are always specified within the font stack itself: `font-family: "Segoe UI", "Segoe UI Web (West European)", ...`
- The base font mixins (e.g. `ms-font-m`) no longer set a `font-family`, as this is inherited from the root `ms-Fabric` class.
- Font family variables have been renamed for clarity, mostly from "font-stack" to "font-family".

# Documentation updates
- Added an example of the `ms-Fabric--selawik` modifier class, which applies a `font-family` that loads the [Selawik web font](https://github.com/Microsoft/Selawik) if Segoe UI is not available locally.
- Shows how to write a custom class that will set the base font family for the application, if another font is preferred to Segoe UI and Selawik. Fixes #871.
- Removed all `font-family` styles and instead sets this through the `ms-Fabric` class applied to the body of the page.
- Added truncation to documentation tables.
- Updated the font weight model to correctly render the table (name and class were reversed).
- Updated the localization table with the correct `lang` attributes, font families, and new samples of text in each language.


mikewheaton:
Marked as "waiting-to-merge", as the Selawik web font has not yet been published to the CDN.

Jahnp:
Approved

This is *awesome* work @mikewheaton!! Super excited for these improvements--Fabric's support for non-West European languages has left much to be desired for quite some time now. With this PR, that changes from "opt-in" afterthought to first-class citizen.

mikewheaton:
Selawik is now on the CDN! Removing the 'waiting-to-merge' label.

micahgodbolt:
Approved

PR #901, opened 2017-01-10
Remove input-placeholder mixin

### LINKED ISSUE #896: Remove mixin: ms-input-placeholder()
`input-placeholder` is [fully implemented](http://caniuse.com/#search=input-placeholder) in the browsers we support, so the `ms-input-placeholder()` mixin is no longer necessary. Even if prefixes were required, that is best left to Autoprefixer rather than a mixin.

mikewheaton:
Fixes #896


micahgodbolt:
Approved

Jahnp:
Approved

PR #902, opened 2017-01-10
Remove utility prefixes

### LINKED ISSUE #882: Remove the distinction of some classes being utilities?
When we adopted the [SUIT CSS naming conventions](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md) early on in Fabric's development, we left many of the classes that are [utilities](https://github.com/suitcss/suit/blob/master/doc/utilities.md) without the `u-` prefix (e.g. `ms-bgColor-yellow`). This was done to prevent breaking changes, while we introduced the `u-` prefix for newer classes like animations and responsiveness (e.g. `ms-u-slideLeftIn10`).

To make this consistent we could either:

1. Follow suit by following SUIT and use the `u-` prefix on most of our classes. These classes would more clearly stand out against component classes (e.g. `ms-Grid` vs `ms-u-sm12`). Any class that is not a component would be a utility.

   Example classes: `ms-u-bgColor-yellow`, `ms-u-textAlignLeft`, `ms-u-slideLeftIn10`, `ms-u-marginLeft`.

   Breaking changes: Class names change for colors and typography.

2. Drop the concept of utilities. It could be argued that everything in Fabric Core is a utility (components are separate and have their own naming convention) and so the distinction isn't helpful here. Removing the `u-` prefix also shortens the class names.

   Example classes: `ms-bgColor-yellow`, `ms-textAlignLeft`, `ms-slideLeftIn10`, `ms-marginLeft`.

   Breaking changes: Class names change for animations, responsiveness, and previous classes categorized as utilities.

Thoughts?

mikewheaton:
Fixes #882

This PR removes the "u-" prefix that denoted many classes as utilities. Given that Fabric is almost entirely utilities, this distinction added more clutter than clarity. This is a breaking change that will require developers to find instances of `ms-u-` and replace them with `ms-`.

Before:
`ms-u-sm12`
`ms-u-slideRightIn10`
`ms-u-borderBase`

After:
`ms-sm12`
`ms-slideRightIn10`
`ms-borderBase`


micahgodbolt:
Approved

Jahnp:
Approved 😀

PR #906, opened 2017-01-19
Add a mixin to ensure font family is inherited

mikewheaton:
Adds a new mixin, `@ms-inherit-font-family`, which sets `font-family: inherit` on button and input elements. This is essentially a very minimal [CSS normalize](http://necolas.github.io/normalize.css/). We can add more elements to it over time as needed. It's used only within the `ms-Fabric` wrapper, ensuring that we don't break any styles outside of our boundary.

The need for this was discovered while doing a test build of Fabric React with the changes that will be included in the next release of Fabric Core. The font-family wasn't being correctly inherited into buttons and form fields because the browser's default styles took precedence.

Before:
image

After:
image

### DIFF
--- a/src/sass/mixins/_General.Mixins.scss
+++ b/src/sass/mixins/_General.Mixins.scss
@@ -244,8 +244,18 @@
 @mixin ms-Fabric {
   -moz-osx-font-smoothing: grayscale;
   -webkit-font-smoothing: antialiased;
+  @include ms-inherit-font-family();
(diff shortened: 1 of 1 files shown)


Jahnp:
Approved

Love it. Makes total sense. This is exactly what we have the `.ms-Fabric` wrapper for--to start setting opt-in, app-wide optimizations. Thanks @mikewheaton!

micahgodbolt:
Approved

PR #911, opened 2017-02-01
Update to the latest value for theme light

mikewheaton:
It turns out that the spec we received way back had an incorrect color in it. This PR updates "theme light" to the correct color value.

color-change


Jahnp:
Approved

micahgodbolt:
Approved

PR #913, opened 2017-02-02
Set '-ms-high-contrast-adjust' to 'none' in the wrapper class

### LINKED ISSUE #907: Set -ms-high-contrast-adjust within ms-Fabric wrapper
Edge has a new feature that sets a black background color for all text in high contrast mode. While this is generally a helpful thing to do, in Fabric we apply our own high contrast styles that can conflict with the default background color.

To prevent Edge from interfering in our high contrast styles, we can add `-ms-high-contrast-adjust:none;` to the ms-Fabric wrapper class.

mikewheaton:
Fixes #907 by applying `-ms-high-contrast-adjust: none;` to the `ms-Fabric` wrapper class.


micahgodbolt:
Approved

Jahnp:
Approved

PR #914, opened 2017-02-02
Add ms-focus-outline() mixin and create variables for focus border style

### LINKED ISSUE #892: Move focus-outline() mixin from Fabric React to Fabric Core
The `focus-outline()` mixin was added to Fabric React in [PR 741](https://github.com/OfficeDev/office-ui-fabric-react/pull/741). This should be in Fabric Core so that other projects can use it.

While doing this, make sure that `ms-focus-outline()` and `ms-focus-border()` use the same border style and color (shared variable would be useful).

mikewheaton:
Fixes #892 

This PR adds a new mixin, `ms-focus-outline()`, which was originally in Fabric React. It also creates a new group of variables for the default width, style, and color that is used by both the `ms-focus-outline()` and `ms-focus-border()` mixins. There are no breaking changes here.


micahgodbolt:
Approved

Jahnp:
Approved

PR #947, opened 2017-04-04
Update ms-background-gradient() mixin and remove unnecessary prefixes

### LINKED ISSUE #904: Update ms-background-gradient() mixin to correct syntax
The [ms-background-gradient() mixin](https://github.com/OfficeDev/office-ui-fabric-core/blob/master/src/sass/mixins/_General.Mixins.scss#L20) currently outputs:

`background-image: linear-gradient(left, ...`

When the [correct syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/linear-gradient) would be:

`background-image: linear-gradient(to left, ...`

I propose we keep passing in a simple string ("left", "right", "top left", etc.) and then use the mixin to add the "to".

### LINKED ISSUE #905: Remove 'ms' prefix from local variables in mixins
Many of the mixins use the `ms-` prefix for local variables, like so:

```
// Drop shadow.
@mixin ms-drop-shadow($ms-x-offset: 0, $ms-y-offset: 0, $ms-blur: 5px, $ms-spread: 0, $ms-alpha: 0.4) {
  box-shadow: $ms-x-offset $ms-y-offset $ms-blur $ms-spread rgba(0, 0, 0, $ms-alpha);
}
```

This is unnecessary, as the prefix is there to prevent conflicts with other global variables. Removing these won't affect the usage of mixins or the output.

mikewheaton:
Fixes #905 by removing unnecessary 'ms-' prefixes from local variable names. This is not a breaking change, as the variables are only referenced within the mixins.

Fixes #904 by updating the `ms-background-gradient()` mixin to output the [proper syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/linear-gradient). Some logic has been added to preserve the existing parameters, so that this is not a breaking change.


battletoilet:
Approved

Jahnp:
Approved

mikewheaton:
#949 should fix the failing build here.

PR #948, opened 2017-04-04
Set the default/base font size in the ms-Fabric wrapper

### LINKED ISSUE #938: Set base font-size in ms-Fabric wrapper
We commonly see `<body class="ms-font-m">` as a means of setting the font size for the entire application. It would make sense to set the font size to 14px within the `.ms-Fabric` wrapper class, so that it's set for the page. This is particularly important for the components, which might scale larger or smaller than expected if the font size is set to something else.

mikewheaton:
Fixes #938 by setting the font size to 14px (medium) in the `ms-Fabric` wrapper component. This is **potentially a breaking change** if an app was previously inheriting font size from a parent element. We should mark it as a breaking change in the release of Fabric Core 7.0.0 to bring awareness to this issue.

### DIFF
--- a/src/sass/mixins/_General.Mixins.scss
+++ b/src/sass/mixins/_General.Mixins.scss
@@ -248,6 +248,7 @@
   @include ms-inherit-font-family();
   color: $ms-color-neutralPrimary;
   font-family: $ms-font-family-west-european;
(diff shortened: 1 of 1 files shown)


battletoilet:
Approved

Agreed that this should go into 7.0.0

PR #950, opened 2017-04-05
Remove unused and undocumented reset mixins

### LINKED ISSUE #895: Remove the 'reset' mixins
We have a whole bunch of mixins in [_General.Mixins.scss](https://github.com/OfficeDev/office-ui-fabric-core/blob/master/src/sass/mixins/_General.Mixins.scss) that are used to set various properties to `normal` or `inherit`. We've never documented these and it looks like only two of them are in use in Fabric React.

Where did these come from? Are they useful? Should we be creating classes (e.g. `ms-resetTransition`) for these and documenting them?

Many of these mixins also contain browser prefixes like `-webkit-transform` that are no longer necessary for the browsers we support. Even if they were, we already have Autoprefixer adding these to our CSS as necessary. Developers who are using these mixins are likely to be supporting different browsers, so the job of providing browser prefixes is best left to their build system or another SCSS mixin library.

mikewheaton:
Fixes #895. Removes various reset mixins that were never documented and aren't widely used.


Jahnp:
Approved

PR #954, opened 2017-04-08
Use font weight variables in the font-face definition mixin

### LINKED ISSUE #953: Semilight font weight set incorrectly in font-face definition
The mixin that generates font-face definitions has a bug, where it specifies that Semilight is `font-weight: 200`. In the classes we use 300. I don't know that either is any more correct than the other, but they do need to match. Let's update the [mixins](https://github.com/OfficeDev/office-ui-fabric-core/blob/201c032f77950ae50f8fd00f4c43591c879e1bc8/src/sass/mixins/_Font.Mixins.scss#L26) to use the SCSS variables for font weight.

Thanks to Eric Porter on Twitter for [pointing out this issue](https://twitter.com/ericporter/status/850514908953366528).

mikewheaton:
Fixes #953 by using the font weight variables in the font-face definitions, ensuring these match everywhere.


Jahnp:
Approved

PR #956, opened 2017-04-10
Add the current icon font version to the CDN URLs

### LINKED ISSUE #955: CDN: Icon font cached too aggressively
The CDN serves a cached version of the icon font to improve performance. However, this cache is not being invalidated by changes to the fonts. When we release an updated font with new icons (as will happen in the next release with PR #946) those icons won't be reliably available. Most, if not all, people will receive the old version instead for an unknown time.

For the Fabric website, we've started using query strings on the address to force the CDN to serve up the latest file. Let's do the same for the icon font, including the font version number like so:

https://static2.sharepointonline.com/files/fabric/assets/icons/fabricmdl2icons.woff2?2.17

This way we can ensure that the latest version of the icon font is served while taking advantage of caching as much as possible.

mikewheaton:
Fixes #955 by including the current icon font version as a query parameter. This will force the CDN to grab the 2.17 version of the font files, rather than serving up a cached copy of the previous version. Going forward, we should update these URLs whenever the icon font is updated.

### DIFF
--- a/src/sass/_Icon.Definitions.scss
+++ b/src/sass/_Icon.Definitions.scss
@@ -13,6 +13,7 @@
 @font-face {
   $icon-font-family: 'FabricMDL2Icons';
   $icon-font-name: 'fabricmdl2icons';
+  $icon-font-version: '2.17';
 
   @if variable_exists(do-scope-styles) {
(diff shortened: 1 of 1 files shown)


Jahnp:
Approved 🍨

Jahnp:
Still approved

PR #959, opened 2017-04-11
Use the correct keyframe names in animation mixins

### LINKED ISSUE #958: Animations are not working
The animation classes refer to animation names like "fadeOut" while the animation keyframes are defined like "ms-fadeOut". This means that none of the animations are working.

mikewheaton:
Fixes #958 by updating the mixins to use the correct, `ms-` prefixed keyframe names.


Jahnp:
Approved

PR #961, opened 2017-04-11
Define a local web font and prevent Chrome/Opera from using it

### LINKED ISSUE #960: Local Segoe UI Semilight not working on Chrome
On Windows systems where the Segoe UI font is installed locally, we prefer using it to downloading the web fonts for performance reasons. However, we've recently noticed that Segoe UI Semilight isn't working correctly in Chrome. See [this CodePen demo](http://codepen.io/mikewheaton/pen/WjezMg?editors=1100) for an example that can be tested across browsers.

I've filed [bug 710539 on Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=710539#) in the hopes that this can be fixed, as it's working in all of the other major browsers. In the meantime, the options I see are to:

- Leave it as is. We get a performance boost on Windows, but we lose the Semilight weight on Chrome.
- Go back to always using web fonts. There's a slight performance penalty (the same as we have on all non-Windows systems) but we have all of the weights available consistently.

Thoughts?

mikewheaton:
Fixes #960 by replacing 'Segoe UI' with a new web font, 'Segoe UI Local', that is defined to only use local fonts. This fixes a bug in IE/Edge where the correct weight was not being applied for Semilight. It also adds an override that targets Chrome and Opera, making those browsers use the web font even where the local font is available. This fixes a bug where those browsers always showed Light instead of Semilight.

See [this CodePen](http://codepen.io/mikewheaton/pen/RVbygB?editors=1100) for a simplified example.

Behavior across browsers:
- macOS: Safari (web font, override so doesn't attempt local)
- macOS: Chrome (web font, override so doesn't attempt local)
- macOS: Firefox (web font, fallback after attempting local)
- Windows 10: Edge (local font)
- Windows 10: IE11 (local font)
- Windows 10: Firefox (local font)
- Windows 10: Chrome (web font, override so doesn't attempt local)
- Windows 10: Opera (web font, override so doesn't attempt local)


Jahnp:
Approved

Cool solution--thanks for demoing the behavior.

mikewheaton:
We've decided to go with the approach in #966 for now, as it's the safest way to ensure that the correct fonts are shown. We'll be investigating methods of loading local fonts in our products and will bring that functionality back to Fabric when it's been tested thoroughly in production.

While this approach worked in simple scenarios, we're concerned that it will conflict with others like language-specific font stacks and custom font-family overrides because the CSS hack relies on specificity. We also haven't had a chance to thoroughly test this approach in real products, and we'd rather err on the side of caution (even if performance takes a small hit) than risk breaking basic functionality like showing the correct font weight.

PR #966, opened 2017-04-13
Remove support for local Segoe UI

### LINKED ISSUE #960: Local Segoe UI Semilight not working on Chrome

mikewheaton:
Fixes #960 by removing local Segoe UI from the font stacks altogether. This means that the Segoe UI web fonts will always be used, even if Segoe UI is installed locally. It's less performant but more reliable, and we can re-introduce support for local Segoe UI in the future when the [Chromium bug] is fixed.

This is an alternative to #961, which uses a browser hack to prevent Chromium browsers from using Segoe UI. My concern is that this may be fragile, and that it hasn't been tested thoroughly in scenarios like language overrides or when developers specify a custom font.


mikewheaton:
We've decided to go this route for now, as it's the safest way to ensure that the correct fonts are shown. We'll be investigating methods of loading local fonts in our products and will bring that functionality back to Fabric when it's been tested thoroughly in production.

mikewheaton:
Reviewed in person with @Jahnp.

PR #977, opened 2017-04-26
Add support for the Lao language

mikewheaton:
This PR adds support for Lao, so that setting an attribute of `lang='lo'` uses a font family of `'Leelawadee UI Web', 'Lao UI', DokChampa, [fallbacks]`.

Example from the docs:

image


Jahnp:
Approved

micahgodbolt:
Approved

PR #988, opened 2017-06-09
Revert -ms-high-contrast-adjust: none in fabric core 7.0.0

mdahamiwal:
This breaks the fabric high contrast for IE and Edge. We need to apply this on selected elements not on overall fabric wrapper.


mdahamiwal:
@mikewheaton, this need to go in soon as it is breaking fabric react high contrast styling for Edge and IE. As per recommendation from Edge team we need to apply this selectively to elements where we control the full high contrast styling. If we apply to entire fabric wrapper everything under it will appear in normal mode since Edge and IE will not map high contrast scheme to colors.

mikewheaton:
@mdahamiwal Thank you for fixing this. This will be included in Fabric Core 7.1.0, which I'll release today.

PR #995, opened 2017-06-30
Remove PullApprove

mikewheaton:
We're now using GitHub's approval workflow.


Jahnp:
Appr... wait a minute...

PR #1029, opened 2017-09-28
Output LTR and RTL styles together

### LINKED ISSUE #1028: Use directionality mixins for RTL, rather than producing a separate stylesheet
### Generating the RTL styles
We currently produce right-to-left (RTL) styles by:
1. Creating a copy of all Fabric Core styles (fabric.css).
2. Adding overrides for icons and animations.
    - Directional icons are swapped (e.g. the `.ms-Icon--Forward` class is produced by the `ms-Icon--Back` mixin).
    - Animations are swapped (e.g. the `.ms-slideLeftIn10` class is produced by the `ms-slideRightIn10` mixin).
3. Running the output through [gulp-css-flipper](https://www.npmjs.com/package/gulp-css-flipper) to replace left with right. For example, the Grid component will place the first column at the right rather than the left.

### Consuming the RTL styles
When using Fabric Core on a page that is being displayed to a right-to-left reading user, the app is responsible for:
1. Referencing `fabric.rtl.css` instead of `fabric.css` to get the correct styles. This can be done on the server for server-rendered pages, or on the client-side via JavaScript.
2. Setting `<html dir='rtl' ... >` to tell the browser to align text to the right and make other adjustments for right-to-left readers.

### Generating the RTL styles
Fabric Core already contains a large set of [directionality mixins](https://github.com/OfficeDev/office-ui-fabric-core/blob/master/src/sass/mixins/_Directionality.Mixins.scss) that output separate style blocks for LTR and RTL readers. These are used wherever the CSS will produce a directional style. For example:

SCSS input:
```
div {
  @include ms-padding-left(20px);
}
```

CSS output:
```
[dir='ltr'] div {
  padding-left: 20px;
}
[dir='rtl'] div {
  padding-right: 20px;
}
```

### Consuming the RTL styles
1. ~~Referencing `fabric.rtl.css` instead of `fabric.css` to get the correct styles. This can be done on the server for server-rendered pages, or on the client-side via JavaScript.~~ There is no need for any additional work, as the styles for both directions are always available.
2. Setting `<html dir='rtl' ... >` to tell the browser to align text to the right and make other adjustments for right-to-left readers.

This is the approach used in [Fabric React](https://github.com/OfficeDev/office-ui-fabric-react) as well as many internal Microsoft applications. While it does slightly increase the file size and can create specificity issues when overriding styles, it's much easier to use. It looks like very few developers are currently supporting RTL, with only 80 CDN requests per month compared to 2.3 million for LTR. Let's reduce the development barriers as much as we can to improve the experience for RTL users.

mikewheaton:
This PR fixes #1028 by using the existing [directionality mixins](https://github.com/OfficeDev/office-ui-fabric-core/blob/master/src/sass/mixins/_Directionality.Mixins.scss) to output separate style blocks for LTR and RTL readers. These are used wherever the CSS will produce a directional style. For example:

SCSS input:
```
div {
  @include ms-padding-left(20px);
}
```

CSS output:
```
[dir='ltr'] div {
  padding-left: 20px;
}
[dir='rtl'] div {
  padding-right: 20px;
}
```

This replaces the previous method of supporting RTL, which relied on developers swapping in an alternate stylesheet for RTL users. That was cumbersome and had very low adoption.

## Breaking changes
- The stylesheets `fabric.rtl.css` and `fabric.rtl.min.css` have been removed. It's no longer necessary to have logic that conditionally loads these, as all styles are contained in `fabric.css` or `fabric.min.css`.
- Even if your app doesn't support RTL users, you must add `dir='rtl'` to the `html` element. This causes the LTR-specific styles to be applied.

## Changes
- All mixins that produce horizontally-asymmetrical styles now use the directionality mixins to target LTR and RTL specifically. For example:
    - `ms-Grid-col` floats to the left side on LTR, right side on RTL
    - `ms-textAlignLeft` will align text to the left on LTR, right on RTL
    - `ms-slideRightIn10` will slide right in LTR, left in RTL
- Updated the internal documentation to add `dir='ltr'` to the `html` element.


mikewheaton:
@lynamemi: I forgot about the OneNote; I'll get that updated tomorrow. 😄

jcupps:
`you must add dir='rtl' to the html element. This causes the LTR-specific styles to be applied.`
I think you mean "you must add dir='ltr'"

Also, not sure if this is the right place for this but it was really confusing to have some icons just disappear with this version. I suggest making 'ltr' the default (i.e. when no direction is specified) with the ability to override it with dir='rtl', especially since dir isn't used all that frequently. Thanks!

PR #1070, opened 2017-11-23
Disable animations based on user preference

### LINKED ISSUE #1068: Disable animations if user prefers no motion
See https://css-tricks.com/introduction-reduced-motion-media-query/

mikewheaton:
Fixes #1068 by wrapping all `animation` and `transition` properties output by the animation classes with the `prefers-reduced-motion: no-preference` media query. This media query will have no impact for the majority of users, where it will be ignored. For those who have opted to 'reduce motion' within the accessibility settings for macOS and iOS, and are using Safari, the animations will be removed so that changes are instant. Here is a [CodePen to test the functionality](https://codepen.io/mikewheaton/pen/aVGGBG?editors=1100).


src/sass/mixins/_Animation.Mixins.scss
```diff
@@ -370,70 +390,55 @@
   }
 }
 
+// @mixin ms-animation($names, $duration, $timing: $ms-animation-ease-1, $fillMode: both) {
```
Jahnp: Remove?

mikewheaton: Whoops, thanks!

PR #1074, opened 2017-12-04
Update icons to version 2.58

### LINKED ISSUE #1073: Update icons to version 2.58
Version 2.58 of the icons is now available.

Jahnp:
Fixes #1073 by updating the icons to version 2.58.

# Deprecated
- HumanResources has been deprecated and will be removed in a future major release.

# New icons
- AnalyticsView
- BrowserScreenShot
- DOM
- HandsFree
- Toll


src/data/icons.json
```diff
@@ -653,7 +657,6 @@
   { "name": "Hospital", "unicode": "E91D" },
   { "name": "Hotel", "unicode": "E824" },
   { "name": "HourGlass", "unicode": "EA03" },
-  { "name": "HumanResources", "unicode": "F240" },
```
mikewheaton: This will remove the icon from the documentation, right? That's probably for the best, as we don't have a way to indicate deprecated icons in the documentation and would prefer that no one starts using it now.

Jahnp: Yeah, removing icons from the JSON will hide them on the site. I think it's the right thing to do in the absence of documenting deprecated icons--though I wonder if we should look into supporting those at some point as well (e.g. with something like `deprecated-icons.json` or something along those lines).

Issue #1088, opened 2018-01-05
Update file type icons to include new and updated symbols


Jahnp:
If you've used OneDrive or a SharePoint Document Library recently, you'll notice that the file type icons have been refreshed with an updated look. 

Take the Excel .xslx icon, for example. Here's the [current icon](https://static2.sharepointonline.com/files/fabric/assets/brand-icons/document/svg/xlsx_48x1.svg) (note that the visual artifact under the "X" does not show up at its intended size of 48px):
![image](https://user-images.githubusercontent.com/307118/34627941-653f67c4-f217-11e7-9c78-9965708f24c9.png)

Here's the [new icon](https://spoprod-a.akamaihd.net/files/fabric/assets/item-types/48/xlsx.svg):
![image](https://user-images.githubusercontent.com/307118/34627948-74ed73aa-f217-11e7-8628-017b5789819a.png)

These icons are actually available today via the [@uifabric/file-type-icons package](https://www.npmjs.com/package/@uifabric/file-type-icons) (the source for which is available in [office-ui-fabric-react](https://github.com/OfficeDev/office-ui-fabric-react/tree/master/packages/file-type-icons), which maps file extensions like .txt and .png to their appropriate PNG and SVG assets. However, these are neither available in Fabric Core, nor are they documented in the Brand Icons page of the Fabric website. 

Fabric Core should be updated to include the new set of icons, its documentation page should be updated to include the full list of file types and sizes, and the docs should include good examples of how to use them.

PR #1090, opened 2018-01-09
BrandIcon: Update to latest document icons

mikewheaton:
Partially addresses #1088 by updating the `ms-BrandIcon` component to support the latest set of document icons. You can view the previous, updated, and new icons in [this CodePen](https://codepen.io/mikewheaton/pen/ZvrvMW).

## Deprecated icons
The `onepkg` and `xls` document icons do not exist in the new set. These have been temporarily mapped to the `one` and `xls` icons to match the [mappings in Fabric React](https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/file-type-icons/src/FileTypeIconMap.ts). These classes will be removed in a future release.

## New icons
24 new product icons have been added:
- archive
- audio
- code
- docset
- email
- exe
- folder
- font
- genericfile
- html
- link
- listitem
- model
- pdf
- photo
- rtf
- sharedfolder
- spo
- sysfile
- txt
- vector
- video
- xml
- zip

### DIFF
--- a/src/sass/_BrandIcon.scss
+++ b/src/sass/_BrandIcon.scss
@@ -1,19 +1,19 @@
 // Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE in the project root for license information.
 
-@include ms-brand-icon-classes(1);
+@include ms-brand-icon-classes('1');
 @include ms-brand-icon-size-classes();
 
 // Pixel Ratio 1.5
 @media only screen and (min-resolution: 144dpi) {
-  @include ms-brand-icon-classes(1_5);
+  @include ms-brand-icon-classes('1_5');
 }
 
 // Pixel Ratio 2
 @media only screen and (min-resolution: 192dpi)  {
-  @include ms-brand-icon-classes(2);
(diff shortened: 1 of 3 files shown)


src/sass/_BrandIcon.scss
```diff
@@ -1,19 +1,108 @@
...
-}
\ No newline at end of file
+  @include ms-brand-icon-classes('3');
+}
+
+// The 'onepkg' and 'xls' icons have been deprecated. These classes will be removed in
+// a future release. Until then, they are mapped to 'one' and 'xlsx'.
+
+.ms-BrandIcon--onepkg.ms-BrandIcon--icon16 {
+  background-image: url(https://spoprod-a.akamaihd.net/files/fabric/assets/item-types/16/one.png);
```
Jahnp: nit: the protocol/domain part of the URL could be pulled into something like a `baseUrl` variable to minimize repetition.

mikewheaton: I'll leave it as is, since there shouldn't be any changes needed here before we remove it entirely.

PR #1109, opened 2018-03-21
Add Fluent grays

mikewheaton:
Adds the new gray palette for the [Fluent Design System](https://fluent.microsoft.com/) updates that are coming to Fabric. This PR includes new variables, mixins, and color classes.

Existing MDL2 colors have been mapped to their equivalents in the new Fluent palette. For example, NeutralSecondary maps to Gray130. The full mapping can be found in `Color.Variable.scss`.

To maintain backward compatibility, these have been added *in addition* to the current color palettes. Eventually, the old colors will be deprecated and removed. To facilitate this future removal, they have been moved to files named like `Color.Variables.MDL2.scss`.

image

### DIFF
--- a/src/documentation/templates/models/NeutralColorsModel.json
+++ b/src/documentation/templates/models/NeutralColorsModel.json
@@ -5,198 +5,184 @@
       "hex": "#000000",
       "fontClass": "ms-fontColor-black",
       "bgClass": "ms-bgColor-black",
-      "borderClass": "ms-borderColor-black",
-      "contrastBgs": [
-        {
-          "class": "ms-bgColor-neutralLight",
-          "ratio": "17:5:1"
-        },
-        {
-          "class": "ms-bgColor-neutralLighter",
-          "ratio": "19:9:1"
-        },
-        {
-          "class": "ms-bgColor-neutralLightAlt",
-          "ratio": "19:8:1"
-        },
-        {
-          "class": "ms-bgColor-themeLight",
-          "ratio": "15:4:1"
-        },
-        {
-          "class": "ms-bgColor-themeLighter",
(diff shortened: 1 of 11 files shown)


src/documentation/templates/models/NeutralColorsModel.json
```diff
@@ -5,198 +5,184 @@
...
-      "secondBgClass": "ms-bgColor-neutralPrimary"
-    },
-    {
-      "name": "White",
-      "hex": "#FFFFFF",
-      "customNameColor": "ms-fontColor-black",
+      "borderClass": "ms-borderColor-black"
+    },
+    {
+      "name": "Gray220",
```
Jahnp: Just thinking out loud--should there be a `legacyName` value or something like that capturing the old name these map to?

mikewheaton: This file is only used for the internal documentation site, which I figured I should update but I don't think anyone is looking at. I'll add an `MDL2equivalent` property to each color that has one, and then we'll have the option to include this in the documentation if we want to.

Jahnp: 👍

mdtauk:
I'm curious as to the thinking behind this palette of Grey colours.  Are these shades going to be replacing the pure Black, White, and Grey/Transparent shades that make up the Windows XAML/UWP Light and Dark theme resources?

Is there a a benefit beyond the aesthetic to using grey shades with varying degrees of R G B values?

Is there a central repository/document which outlines core Fluent changes which Fabric and potentially other implementations of Fluent will align to?

Does this apply to control designs, which currently vary between Fluent UWP, Fabric UI, Fluent Web etc?

PR #1114, opened 2018-04-09
Add main color palette

mikewheaton:
Adds the main (illustration) color set variables, mixins, and classes. Where possible, old color values have been mapped to the new color set. All of the previous colors have been marked as deprecated but have not been removed.


mikewheaton:
@lynamemi Purple is probably a more descriptive/common name to use. The names come from the color wheel segments shown below, so blue-magenta is used for consistency since it falls between blue and magenta.

image

PR #1140, opened 2018-09-04
Fluent: Add motion variables, mixins, and classes

mikewheaton:
Adds variables, mixins, and classes for the new Fluent motions. All of the existing animation variables, mixins, and classes have been moved to files marked `MDL2` for deprecation. These are imported by the new files to preserve compatibility.

Note that the new items are named 'motion' rather than 'animation'. This matches the language used by the designers and allows MDL2 and Fluent to coexist without naming conflicts.


src/sass/mixins/_Motion.Mixins.scss
```diff
@@ -0,0 +1,204 @@
...
+}
+
+// Slides.
+@mixin ms-motion-slideLeftOut {
+  @include ms-LTR {
+    animation: ms-motion-slideLeftOut 100ms $ms-motion-timing-accelerate both,
+      ms-motion-fadeOut 100ms $ms-motion-timing-linear both;
+  }
+
+  @include ms-RTL {
```
mikewheaton: For horizontal slides, the direction is reversed for right-to-left (RTL).

Jahnp:
Just a general thought: There are a few renames here (Animation --> Motion, Animation --> Animation.MDL2) that will break downstream consumers if they're looking for those files. Even though we don't document them explicitly, we know they're often used this way. We should be careful about renames on branches we know are in active use.

We should consider re-exporting the new variable names from files with the old names (e.g. retaining the `_Animation` files) until we've done a major release containing the change.

mikewheaton:
I've renamed the files from Motion to Animation to maintain compatibility for projects that are importing partials rather than using References.scss.

src/sass/variables/_Animation.Variables.MDL2.scss
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE in the project root for license information.
+
```
Jahnp: I know it's implied by the filename, but we should add a comment in the file that these are now deprecated

mikewheaton: Added

PR #1142, opened 2018-09-11
Fluent: Update to the latest Shared color palette

mikewheaton:
Updates the Shared color palette to the latest colors.

### DIFF
--- a/src/sass/variables/_Color.Variables.scss
+++ b/src/sass/variables/_Color.Variables.scss
@@ -38,8 +38,9 @@ $ms-color-white: #ffffff !default;
 
 // Shared
 $ms-color-sharedPinkRed10: #750b1c !default;
-$ms-color-sharedRed30: #a4262c !default;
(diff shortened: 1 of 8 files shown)


Jahnp:
These look good to me, but I'd just want to make sure the changes (e.g. `sharedRed30` --> `sharedRed20`) don't impact @leddie24's Fluent work. @leddie24 can you take a look when you have a chance?

src/sass/variables/_Color.Variables.scss
```diff
@@ -38,8 +38,9 @@ $ms-color-white: #ffffff !default;
 
 // Shared
 $ms-color-sharedPinkRed10: #750b1c !default;
-$ms-color-sharedRed30: #a4262c !default;
```
Jahnp: @mikewheaton just confirmed that removing this will break a scenario @leddie24 is working on. Let's work to address this offline before publishing a new version.

PR #1144, opened 2018-09-27
Variable cdn path for icons, as per fonts

thornyweb:
Create and use variable for icons path rather than being hard coded.

Also change a couple of hard coded font paths to use the cdn variable that already exists.


mikewheaton:
The build issue is caused by the Travis config specifying an outdated version of Node. This has been fixed on the Fluent branch (see [this commit](https://github.com/OfficeDev/office-ui-fabric-core/commit/79778740aa6addb71fe7964045952c210eafe7ab#diff-354f30a63fb0907d4ad57269548329e3)) but not yet on master. It can be done in a separate PR.

thornyweb:
Thanks @mikewheaton

PR #1242, opened 2020-11-10
Fix missing edit (and other) icons

sayers24:
Customers have complained that the Edit, Datetime and other icons are missing due to the added [dir='ltr'] selector that may not be in the customer's markup. Removing the scoping selector will allow for use of dir="auto" as well as avoid confusion that something might be broken


sayers24:
Issue: https://github.com/OfficeDev/office-ui-fabric-core/issues/1236

lindalu-MSFT (approved):

PR #1289, opened 2022-12-07
Ran prettier on src and gulp files

micahgodbolt:


micahgodbolt:
I feel that this might be a bit heavy handed. The thought was to run prettier on everything so that future PRs wouldn't have random formatting changes, but not sure if this much churn is worth it

PR #1290, opened 2022-12-07
feat: Add new m365 icon, add brand icons page to docs, and refactor code

sopranopillow:
This PR adds the following:
* Adds test, start, and build commands.
* Adds a new section to the docs to visualize the brand icons.
* Refactors cdn url usage and updates to latest cdn link.
* Adds new m365 icons.

Fixes https://github.com/microsoft/fluentui/issues/25928

### DIFF
--- a/src/sass/_BrandIcon.scss
+++ b/src/sass/_BrandIcon.scss
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE in the project root for license information.
 
+@import './variables/_Icon.Variables.scss';
 @include ms-brand-icon-classes('1');
 @include ms-brand-icon-size-classes();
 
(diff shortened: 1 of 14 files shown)


src/sass/variables/_Icon.Variables.scss
```diff
@@ -1,7 +1,9 @@
 // Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE in the project root for license information.
 
 // Icon directory and paths
-$ms-icon-cdn-path: 'https://static2.sharepointonline.com/files/fabric/assets/icons' !default;
+$ms-cdn-build-date: '20221201.001';
+$ms-cdn-path: 'https://res-1.cdn.office.net/files/fabric-cdn-prod_#{$ms-cdn-build-date}/assets' !default;
```
micahgodbolt, line 5: lets move these out to another file. i think this one should only be about icons in the same way that BrandIcons.Variables is only about BrandIcons

maybe keep it simple with a _cdn.variables.scss file

src/sass/_BrandIcon.scss
```diff
@@ -22,87 +23,87 @@
 // a future release. Until then, they are mapped to 'one' and 'xlsx'.
 
 .ms-BrandIcon--onepkg.ms-BrandIcon--icon16 {
-  background-image: url(https://spoprod-a.akamaihd.net/files/fabric/assets/item-types-fluent/16/one.png);
+  background-image: url('#{$ms-cdn-path}/item-types-fluent/16/one.png');
```
Jahnp, line 26: As long as you're changing these, you should be able to replace `item-types-fluent` with `item-types`. That's the more up to date and maintained assets directory.

Jahnp (approved):
Thanks for this @sopranopillow! Looks great.

src/sass/_BrandIcon.scss
sopranopillow: will make the change!

